PR #4, opened 2022-03-02
Not just another custom column widths PR

Slamdunk:
See https://github.com/box/spout/pull/715


Slamdunk:
Thank you @aphofstede

aphofstede:
No problem. Hope this project gets the maintenance it deserves.

BelegUS:
Are there docs for this feature?

Also, is it really so that column width can be set only **before** any data is written to the spreadsheet? Seems so - and that's kinda bummer, as it doesn't allow setting column widths to adjust them to longest value written to each column. Having to know these values before write results in very unoptimal code when eg. writing data from iterator.

Is it an XLSX limitation?

aphofstede:
See description of the original PR. Since the maintainers had declined multiple PRs already we wanted to run this implementation by them first before documenting and building further on top of this.
If this fork gets much traction we should add docs and revisit an autosizing option.

Slamdunk:
Hi @BelegUS it is my plan to postpone every actual `fwrite` as late as possible.

Problem is that Microsoft Office don't like having `<cols>` section after `<sheetData>` one, even though XML standard allow random order of elements.
So we can't actually write column widths after data processing.

The only solution would be to process data, get the file pointer back to the top and overwrite a dummy portion of the file just like it was proposed here https://github.com/box/spout/pull/866/files#diff-f1797564831f00a171c0864b91771b3ba9a4b1b2c7980c7a91dff07d79f55bc3R130-R134

But that's a nightmare for maintenance reasons, I'm not incline to implement that at the moment.

Slamdunk:
Hi, just like ODS already did it, with https://github.com/openspout/openspout/pull/45 now also XLSX writes the rows in a temporary file and only after that it writes all the remaining data, thanks to the speed of `stream_copy_to_stream`.

So now column with, alongside auto-sizing, could be easily supported and already achievable by user code :rocket:

PR #21, opened 2022-03-10
Add support for mergeCells, shrink to fit

Slamdunk:
See https://github.com/box/spout/pull/738

- https://github.com/box/spout/pull/729 - fix shrink to fit + add test shrinkToFit was not handled in StyleMerger so it was overwritten by the default cell style StyleManager didn't add the property to the xml if shrinkToFit was used without alignment or text wrap. Unit test
- https://github.com/box/spout/pull/529 - tests for column width, row height and cell merge + minor refact Tests Added 'addOption' to OptionsManagerInterface Moved 'setColumnWidths' and 'mergeCells' methods to Xlsx Writer implementation since the actual feature only works for Xlsx at the moment

### DIFF
--- a/src/Common/Entity/Row.php
+++ b/src/Common/Entity/Row.php
@@ -21,7 +21,8 @@ class Row
     protected $style;
 
     /**
-     * Row height (default is 15)
+     * Row height (default is 15).
+     *
      * @var string
      */
     protected $height = '15';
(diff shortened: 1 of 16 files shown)


Slamdunk:
Thank you @quamis and @ignaczistvan

PR #30, opened 2022-03-15
Mark all classes as final

Slamdunk:


Slamdunk:
Hi, thank you for your interest in this project.
About `final`:

1. https://ocramius.github.io/blog/when-to-declare-classes-final/
2. https://en.wikipedia.org/wiki/Composition_over_inheritance
3. More closed the source is to _overwrite_ features like Inheritance, more robust it is to bugs and easier to maintain
4. If everything is open and we need to close it for some reason, it would be a BC break and thus a new major release. Instead, if everything starts closed and a new functionality needs opening a class, we can do it in a minor version.

uncaught:
What was the motivation in making everything final and thus preventing any sort of customization of services? This is after all a library and not an end product.

Slamdunk:
@uncaught https://github.com/openspout/openspout/pull/30#issuecomment-1068778672

PR #43, opened 2022-03-21
Add benchmarks to CI

Slamdunk:


Slamdunk:
@adrilo I've moved the `perf-tests` test groups into [`phpbench/phpbench`](https://github.com/phpbench/phpbench) benchmarks and added them into the CI build, so we can easily track eventual regressions, hope you find it interesting:

image

image

PR #59, opened 2022-03-25
fix: throw exception when cell is passed into another cell

darthf1:
Hi!

This PR is to throw a decent error when a cell is passed to another cell as a value.

I had the following in my code:
```php
$cells = [Cell::fromValue('')];
$row = Row::fromValues($cells);
```

Which lead to the stacktrace:
```
OpenSpout\Common\Exception\InvalidArgumentException : Trying to add a value with an unsupported type: NULL
 /home/www/app/vendor/openspout/openspout/src/Writer/XLSX/Manager/WorksheetManager.php:218
 /home/www/app/vendor/openspout/openspout/src/Writer/XLSX/Manager/WorksheetManager.php:134
 /home/www/app/vendor/openspout/openspout/src/Writer/XLSX/Manager/WorksheetManager.php:95
 /home/www/app/vendor/openspout/openspout/src/Writer/Common/Manager/AbstractWorkbookManager.php:260
 /home/www/app/vendor/openspout/openspout/src/Writer/Common/Manager/AbstractWorkbookManager.php:130
 /home/www/app/vendor/openspout/openspout/src/Writer/AbstractWriterMultiSheets.php:105
 /home/www/app/vendor/openspout/openspout/src/Writer/AbstractWriter.php:107
```

With this PR, it leads to the stacktrace:
```
OpenSpout\Common\Exception\InvalidArgumentException : Cannot pass an instance of OpenSpout\Common\Entity\Cell as a value.
 /home/www/app/vendor/openspout/openspout/src/Common/Entity/Cell.php:44
 /home/www/app/vendor/openspout/openspout/src/Common/Entity/Row.php:40
 /home/www/app/vendor/openspout/openspout/src/Common/Entity/Row.php:39
```

The fix was to do:
```php
$cells = [Cell::fromValue('')];
$row = new Row($cells);
```

PR #60, opened 2022-03-25
Restrict allowed Cell value types to the actual ones

Slamdunk:
See https://github.com/openspout/openspout/pull/59

### DIFF
--- a/src/Common/Entity/Cell.php
+++ b/src/Common/Entity/Cell.php
@@ -10,7 +10,6 @@ use OpenSpout\Common\Entity\Cell\BooleanCell;
 use OpenSpout\Common\Entity\Cell\DateIntervalCell;
 use OpenSpout\Common\Entity\Cell\DateTimeCell;
 use OpenSpout\Common\Entity\Cell\EmptyCell;
-use OpenSpout\Common\Entity\Cell\ErrorCell;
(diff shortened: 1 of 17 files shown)


src/Common/Entity/Cell.php
```diff
@@ -37,7 +36,7 @@ final public function getStyle(): Style
         return $this->style;
     }
 
-    final public static function fromValue(mixed $value, ?Style $style = null): self
+    final public static function fromValue(null|bool|string|int|float|DateTimeInterface|DateInterval $value, ?Style $style = null): self
```
Slamdunk, line 39: I do not consider this a BC break since the library just broke if any other type where used.

src/Writer/AbstractWriter.php
```diff
@@ -103,16 +110,7 @@ final public function addRow(Row $row): void
             throw new WriterNotOpenedException('The writer needs to be opened before adding row.');
         }
 
-        try {
-            $this->addRowToWriter($row);
-        } catch (OpenSpoutException $e) {
-            // if an exception occurs while writing data,
-            // close the writer and remove all files created so far.
-            $this->closeAndAttemptToCleanupAllFiles();
```
Slamdunk, line 111: It turned out that all this code where only involved with invalid Cell value types

PR #75, opened 2022-04-18
Specify worksheet (by index) in cells merge

tntrex:
Currently, if you merge cells on multiple worksheets, all merges will be applied on all worksheets. So tables will look broken in Excel. This is hotfix PR for the problem.

P.S. Sorry about PR without test case. I have no time for it right now.

### DIFF
--- a/src/Writer/XLSX/Options.php
+++ b/src/Writer/XLSX/Options.php
@@ -6,6 +6,8 @@ namespace OpenSpout\Writer\XLSX;
 
 use OpenSpout\Common\Entity\Style\Style;
 use OpenSpout\Writer\Common\AbstractOptions;
+use OpenSpout\Writer\Exception\WriterNotSet;
+use WeakReference;
 
 final class Options extends AbstractOptions
 {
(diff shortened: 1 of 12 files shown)


Slamdunk:
Hi, thank you for your contribution, it looks promising.

Take your time to write tests, there's no hurry 😉

tntrex:
Hello @Slamdunk 

I've just repushed branch: added test case, fixed codestyle and phpstan noticies

src/Writer/XLSX/Options.php
```diff
@@ -43,7 +62,13 @@ public function mergeCells(
         int $bottomRightColumn,
         int $bottomRightRow
     ): void {
+        $writer = $this->getWriter();
```
Slamdunk: Stateful `Options` would be confusing to use and hard to maintain.

We can simply extend `Options::mergeCells` API to have a 5th parameter for the `$sheetIndex`, defaulted to `0` to keep backward compatibility.

Slamdunk (changes requested):
Overall it looks good, just simplify the API and we can :ship:

src/Writer/XLSX/Options.php
tntrex: Done refactoring

Slamdunk (approved):

Slamdunk:
Thank you @hustlahusky

PR #63, opened 2022-03-28
Add setCellVerticalAlignment() & allow setShouldWrapText() to be explicitly set as false

Slamdunk:
See:

1. https://github.com/box/spout/issues/737
1. https://github.com/box/spout/issues/829
1. https://github.com/box/spout/pull/869
1. https://github.com/openspout/openspout/issues/62

PR #77, opened 2022-04-26
[XLSX] Add feature to set row height

jonnott:
I'm trying to add support for setting row height, but I don't know how to do it.

I've re-added getter/setter and a property which were removed in a recent commit due to being unused. I've also added a failing test.

What I don't know is what the syntax in the raw XML output (XLS or ODS) would be for specifying the height of the row, or where this code would need to be.

The code would check for strlen() of the height property, and then apply the height to the row if height has been set via the setter.

I recently did a PR for celllVerticalAlignment and explicit setWrapText to false: https://github.com/openspout/openspout/pull/63

I'm trying to replicate doing something like this equivalent in XLSXWriter:

```
if ($lines > 1) { $row_options['height'] = $lines * 12.1; }
$writer->writeSheetRow('Sheet1', array_values($row), $row_options);
```

Any help or pointers would be much appreciated. I'd be happy to fully finish this PR and tests if only I knew how!

### DIFF
--- a/src/Common/Entity/Row.php
+++ b/src/Common/Entity/Row.php
@@ -20,6 +20,9 @@ final class Row
     /** The row style. */
     private Style $style;
 
+    /** Row height. */
+    private string $height = '';
+
     /**
      * Row constructor.
      *
@@ -106,6 +109,24 @@ final class Row
         $this->style = $style ?? new Style();
 
         return $this;
+    }
+
+    /**
+     * Set row height.
+     */
+    public function setHeight(string $height): self
+    {
+        $this->height = $height;
+
+        return $this;
+    }
+
+    /**
+     * Returns row height.
+     */
+    public function getHeight(): string
+    {
+        return $this->height;
     }
 
     /**
--- a/src/Common/Entity/Row.php
+++ b/src/Common/Entity/Row.php
@@ -21,7 +21,7 @@ final class Row
     private Style $style;
 
     /** Row height. */
-    private string $height = '';
+    private float $height = 0;
 
     /**
      * Row constructor.
@@ -114,7 +114,7 @@ final class Row
     /**
      * Set row height.
      */
-    public function setHeight(string $height): self
+    public function setHeight(float $height): self
     {
         $this->height = $height;
 
@@ -124,7 +124,7 @@ final class Row
     /**
      * Returns row height.
      */
-    public function getHeight(): string
+    public function getHeight(): float
     {
         return $this->height;
     }
--- a/tests/Writer/XLSX/WriterTest.php
+++ b/tests/Writer/XLSX/WriterTest.php
@@ -10,6 +10,7 @@ use DOMElement;
 use finfo;
 use OpenSpout\Common\Entity\Cell;
 use OpenSpout\Common\Entity\Row;
+use OpenSpout\Common\Entity\Style\Style;
 use OpenSpout\Common\Exception\IOException;
 use OpenSpout\Reader\Wrapper\XMLReader;
 use OpenSpout\TestUsingResource;
@@ -429,6 +430,20 @@ final class WriterTest extends TestCase
         $this->writeToXLSXFile($dataRows, $fileName);
 
         $this->assertInlineDataWasWrittenToSheet($fileName, 1, 'control _x0015_ character');
+    }
+
+    public function testAddRowShouldApplyHeight(): void
+    {
+        $fileName = 'test_add_row_should_apply_height.xlsx';
+
+        $this->writeToXLSXFile([Row::fromValues(['xlsx--11'])->setHeight('25')], $fileName);
+
+        $xmlReader = $this->getXmlReaderForSheetFromXmlFile($fileName, '1');
+
+        $xmlReader->readUntilNodeFound('row');
+        $DOMNode = $xmlReader->expand();
+        self::assertEquals('25', $DOMNode->getAttribute('ht'), 'Row height does not equal given value.');
+        self::assertEquals('1', $DOMNode->getAttribute('customHeight'), 'Row does not have custom height flag set.');
     }
 
     public function testCloseShouldAddMergeCellTags(): void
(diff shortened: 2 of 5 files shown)


Slamdunk (changes requested):
Implementing the feature is the easiest part, we need the test first.

> What I don't know is what the syntax in the raw XML output (XLS or ODS) would be for specifying the height of the row

That's pretty easy to look for: both XLSX and ODS are just a `zip` file of XMLs, so:

1. open LibreOffice/MSOffice
2. write a cell
3. save the document as XLSX/ODS
4. change the height of a row
5. save the document as XLSX/ODS with a different filename
6. rename the files appending `.zip` at the end
7. unzip the files
8. use programs like `meld` or `winmerge` to inspect the two folders

You'll exactly see how the custom row height is written

jonnott:
> Implementing the feature is the easiest part, we need the test first.

I've updated with a working implementation and passing test for XLSX. I have *no idea* how to do the equivalent for ODS.

This is the diff between my test XLSX files:

image

..and the diff between my test ODS files:

image

Anyone able to help with the ODS implementation?

jonnott:
@Slamdunk

faytekin:
@Slamdunk

Slamdunk:
@faytekin may you help @jonnott with the ODS implementation?

delameter:
Fixed broken pipeline in 7f8ff57a90b7453135138e4a595df040c69424ed (made a PR to original branch in author's repo).

Also researched a bit about ODS implementation, and it's a lot harder than it seems to be.  I'll cut to the chase, in ODS "height" is the attribute of style, not the attribute of row. I attempted to insert elements describing row style `<style:table-row-properties style:row-height="$$$pt" style:use-optimal-row-height="false"/>` into `table` section before row definition, but they were ignored by Libre (small wonder). Libre itself saves row height like this:

image

But AFAIK writing row styles is not implemented in current version (only cell styles). It's hard to say how long will it take for someone to fully implement this, if it's even possible considering write-only approach.

At the same time, setting the height of XLSX rows seems to work (the only thing that I can't comprehend is why @jonnott used `string` variable to store the height and not `float`), so how about separating these two features and releasing customizable heights for XLSX only?

jonnott:
> (the only thing that I can't comprehend is why @jonnott used `string` variable to store the height and not `float`)

@delameter Feel free to change this if you feel `float` would be better.

jonnott:
> > (the only thing that I can't comprehend is why @jonnott used `string` variable to store the height and not `float`)
> 
> @delameter Feel free to change this if you feel `float` would be better.

I think it was just because the pre-existing `setHeight()` and `getHeight()` (which I reinstated after they had been removed) used `string` for this property.

delameter:
I also made `Dockerfile` and `docker-compose.yml` for fast tests launching and debugging without fiddling with dependencies, but don't know will it fit in or not. If it's acceptable I can create a separate PR for adding Docker support.

delameter:
> > > (the only thing that I can't comprehend is why @jonnott used `string` variable to store the height and not `float`)
> > 
> > 
> > @delameter Feel free to change this if you feel `float` would be better.
> 
> I think it was just because the pre-existing `setHeight()` and `getHeight()` (which I reinstated after they had been removed) used `string` for this property.

Well I can assume that it was initially made for possibility to set the height with units, like "15pt" or "2cm",  but then there should be at least some validation mechanism (even if rudimentary) to ensure that internal file markup will not be broken.

jonnott:
> Well I can assume that it was initially made for possibility to set the height with units, like "15pt" or "2cm",

Yes, I think that's why.

jonnott:
@Slamdunk Any ideas who might be able to help get this one over the line?

src/Common/Entity/Row.php
```diff
@@ -20,6 +20,9 @@ final class Row
     /** The row style. */
     private Style $style;
 
+    /** Row height. */
+    private string $height = '';
```
Slamdunk:
```suggestion
    private float $height = 0;
```

src/Common/Entity/Row.php
```diff
@@ -108,6 +111,24 @@ public function setStyle(?Style $style): self
         return $this;
     }
 
+    /**
+     * Set row height.
+     */
+    public function setHeight(string $height): self
```
Slamdunk:
```suggestion
    public function setHeight(float $height): self
```

src/Common/Entity/Row.php
```diff
@@ -108,6 +111,24 @@ public function setStyle(?Style $style): self
...
+    {
+        $this->height = $height;
+
+        return $this;
+    }
+
+    /**
+     * Returns row height.
+     */
+    public function getHeight(): string
```
Slamdunk:
```suggestion
    public function getHeight(): float
```

Slamdunk (changes requested):
I'm ok to merge this without ODS support.

Let's just use native types and we're ok

jonnott:
Committed those changes @Slamdunk

tests/Writer/XLSX/WriterTest.php
```diff
@@ -430,6 +431,20 @@ public function testAddRowShouldEscapeControlCharacters(): void
         $this->assertInlineDataWasWrittenToSheet($fileName, 1, 'control _x0015_ character');
     }
 
+    public function testAddRowShouldApplyHeight(): void
+    {
+        $fileName = 'test_add_row_should_apply_height.xlsx';
+
+        $this->writeToXLSXFile([Row::fromValues(['xlsx--11'])->setHeight('25')], $fileName);
```
Slamdunk:
```suggestion
        $this->writeToXLSXFile([Row::fromValues(['xlsx--11'])->setHeight(25)], $fileName);
```

Slamdunk (changes requested):
We forgot this one I guess

jonnott:
Fixed test also

Slamdunk (changes requested):
Coding Standards and Static Analysis must be fixed too

jonnott:
@Slamdunk I have no idea why the Static Analysis says "Error: Cannot call method getAttribute() on DOMNode|false.". I therefore cannot implement a solution. I can't see how it's got anything to do with this PR specifically.

Slamdunk (approved):

PR #78, opened 2022-04-29
Writer: add AutoFilter capability

bionda740:


src/Writer/AbstractWriterMultiSheets.php
```diff
@@ -81,6 +82,37 @@ final public function setCurrentSheet(Sheet $sheet): void
...
+    /**
+     * Row coordinates are indexed from 1, columns from 0 (A = 0),
+     * so a filter B2:G2 looks like $writer->setAutoFilterForCurrentSheet(1, 2, 6, 2);.
+     *
+     * @param 0|positive-int $fromColumnIndex
+     * @param positive-int   $fromRow
+     * @param 0|positive-int $toColumnIndex
+     * @param positive-int   $toRow
+     */
+    public function setAutoFilterForCurrentSheet(
```
Slamdunk: No need to add a new API to `AbstractWriterMultiSheets`, adding it to `Sheet` should suffice

src/Writer/Common/Entity/Worksheet.php
```diff
@@ -92,4 +99,17 @@ public function getId(): int
         // sheet index is zero-based, while ID is 1-based
         return $this->externalSheet->getIndex() + 1;
     }
+
+    public function getAutoFilter(): AutoFilter
```
Slamdunk: This API can be moved to `Sheet` which already holds configurations for the actual sheets

src/Writer/XLSX/AutoFilter.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public function setRange(string $range): self
+    {
+        if ('' === $range) {
+            $this->range = '';
+
+            return $this;
+        }
+
+        $preg_match_all = preg_match_all('/([A-Z]+)(\d+):([A-Z]+)(\d+)/', $range, $matches);
```
Slamdunk: To keep consistency with the library, we can just copy-paste what's already done in `MergeCell`, and call `CellHelper::getColumnLettersFromColumnIndex` in the `FileSystemHelper` only.

tests/Writer/XLSX/AutoFilterTest.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Writer\XLSX;
```
Slamdunk: Nice, we also need a tests for `ODS` format

Slamdunk (changes requested):

Slamdunk (approved):

Slamdunk:
I'll take care of `CellValueFormatter::formatExcelTimestampValueAsDateTimeValue` PHPStan error, thank you

PR #83, opened 2022-06-13
RowIteratorInterface: add `null` as a valid return type

jwhulette:
The row buffer can can be null per
`
    /** @var null|Row Buffer used to store the current row, while checking if there are more rows to read */
    private ?Row $rowBuffer;
`

Updated the functions to allow returning null.

### DIFF
--- a/src/Reader/RowIteratorInterface.php
+++ b/src/Reader/RowIteratorInterface.php
@@ -8,9 +8,7 @@ use Iterator;
 use OpenSpout\Common\Entity\Row;
 
 /**
- * @extends Iterator<Row>
- *
- * @template-extends Iterator<int, null|Row>
+ * @extends Iterator<null|Row>
  */
 interface RowIteratorInterface extends Iterator
 {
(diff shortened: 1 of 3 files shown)


Slamdunk:
Hi, can you add a test to prove the need for such change please?

jwhulette:
Added test for row iterator current returning null.

jwhulette:
I'm working on the phpstan error.

Slamdunk:
To develop locally, type `make`

Slamdunk:
Thank you

src/Reader/RowIteratorInterface.php
```diff
@@ -8,9 +8,9 @@
 use OpenSpout\Common\Entity\Row;
 
 /**
- * @extends Iterator<Row>
+ * @extends Iterator<null|Row>
```
stof, line 11: is the Row actually nullable during iteration ? Or can the `current()` function return null only for iterators that are not in the valid state ? If the iteration never returns null, this line should be reverted.

From looking at the code, the rowBuffer cannot be null when the `valid()` method returns `true`.

Slamdunk: You are right, only a raw call to `current()` method can return `null`, the Iterator itself never iterates over `null` values.

Can you pick this fix up please?

stof: Here it is: https://github.com/openspout/openspout/pull/86

PR #86, opened 2022-06-20
Fix the iteration type for RowIteratorInterface

stof:
The current method can only return null when calling it on a non-valid iterator state, which the iteration does not do.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6,6 +6,11 @@ parameters:
 			path: src/Common/Entity/Row.php
 
 		-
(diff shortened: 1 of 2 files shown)


stof:
another option is to make the iterator throw a LogicException when calling `current` without a Row.

Slamdunk:
I prefer this PR to be honest. Manually calling `current()` is legit

stof:
@Slamdunk but phpstan complains about it right now.

And is calling `current` manually legit when the iterator is not in a valid state ? There is no promise about what happen in that case.

Slamdunk:
The `LogicException` would just mimic the current `TypeError` then, I don't see the point of it.

I'd go with suppressing the PHPStan errors: who uses `foreach` or `iterator_to_array` have no `null`s, who uses the `current()` call have to handle the `null` :shrug:

stof:
@Slamdunk I added them in the baseline.

Slamdunk:
Thank you

PR #87, opened 2022-06-20
fix: duplicate xlsx files on Windows

### LINKED ISSUE #85: styles.xml and workbook.xml are added twice to the archive on Windows
Using PHP 8.1.5 and OpenSpout 4.6.2 on Windows 10.

addFileToArchiveWithCompressionMethod() adds "xl\workbook.xml" and "xl\styles.xml"
then addFolderToArchive() re-adds them as "xl/workbook.xml" and "xl/styles.xml"
Notice the different directory separators.

$zip->locateName() in shouldSkipFile() fails to see them as duplicates and they are not overwritten either. The files end up in the zip archive twice. The end result is that at least Excel 2010 says that the file is corrupt (but is able to repair it). And of course the file is bigger than it should be.

I upgraded to OpenSpout from an older Spout 2.x, where this issue didn't exist. So then I noticed that if I revert this change in the FileSystemHelper.php, everything works fine again:

```
$this->zipHelper->addFileToArchive($zip, $this->rootFolder, self::XL_FOLDER_NAME.\DIRECTORY_SEPARATOR.self::WORKBOOK_XML_FILE_NAME);
$this->zipHelper->addFileToArchive($zip, $this->rootFolder, self::XL_FOLDER_NAME.\DIRECTORY_SEPARATOR.self::STYLES_XML_FILE_NAME);
```
=>

```
$this->zipHelper->addFileToArchive($zip, $this->rootFolder, self::XL_FOLDER_NAME . '/' . self::WORKBOOK_XML_FILE_NAME);
$this->zipHelper->addFileToArchive($zip, $this->rootFolder, self::XL_FOLDER_NAME . '/' . self::STYLES_XML_FILE_NAME);
```

sutrik:
fix #85

### DIFF
--- a/src/Writer/Common/Helper/ZipHelper.php
+++ b/src/Writer/Common/Helper/ZipHelper.php
@@ -149,11 +149,12 @@ final class ZipHelper
      */
     private function addFileToArchiveWithCompressionMethod(ZipArchive $zip, string $rootFolderPath, string $localFilePath, string $existingFileMode, int $compressionMethod): void
     {
-        if (!$this->shouldSkipFile($zip, $localFilePath, $existingFileMode)) {
+        $normalizedLocalFilePath = str_replace(\DIRECTORY_SEPARATOR, '/', $localFilePath);
+        if (!$this->shouldSkipFile($zip, $normalizedLocalFilePath, $existingFileMode)) {
             $normalizedFullFilePath = $this->getNormalizedRealPath($rootFolderPath.'/'.$localFilePath);
-            $zip->addFile($normalizedFullFilePath, $localFilePath);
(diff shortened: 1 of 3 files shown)


Slamdunk:
Hi, thank you for the contribution. Can you also add some test for the bug so we are sure it never gets reintroduced in the future?

sutrik:
> Hi, thank you for the contribution. Can you also add some test for the bug so we are sure it never gets reintroduced in the future?

Sure, I'll look into it.

sutrik:
I decided to do the fix in a different way. Otherwise making a test would have required creating interfaces for FileSystemHelper and ZipHelper just so they could have been mocked.

Slamdunk (approved):

Slamdunk:
Thank you @sutrik

PR #113, opened 2022-11-11
Row outline/hidden/collapsed attributes support (xlsx only)

LastDragon-ru:
image

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -218,13 +218,25 @@ $writer->openToFile('/tmp/file.xlsx');
 
 $levelOne = new RowAttributes(1);
 $levelTwo = new RowAttributes(2);
+$rowA     = Row::fromValues([1, 1, 1]);
+$rowB     = Row::fromValues([2, 2, 2]);
+$rowC     = Row::fromValues([3, 3, 3]);
+$rowD     = Row::fromValues([4, 4, 4]);
+$rowE     = Row::fromValues();
+$rowF     = Row::fromValues();
 
-$writer->addRow($options->setRowAttributes(Row::fromValues([1, 1, 1]), $levelOne));
-$writer->addRow($options->setRowAttributes(Row::fromValues([2, 2, 2]), $levelOne));
-$writer->addRow($options->setRowAttributes(Row::fromValues([3, 3, 3]), $levelTwo));
-$writer->addRow($options->setRowAttributes(Row::fromValues([4, 4, 4]), $levelTwo));
-$writer->addRow($options->setRowAttributes(Row::fromValues(), $levelOne));
-$writer->addRow(Row::fromValues());
+$options->setRowAttributes($rowA, $levelOne);
+$options->setRowAttributes($rowB, $levelOne);
+$options->setRowAttributes($rowC, $levelTwo);
+$options->setRowAttributes($rowD, $levelTwo);
+$options->setRowAttributes($rowE, $levelOne);
+
+$writer->addRow($rowA);
+$writer->addRow($rowB);
+$writer->addRow($rowC);
+$writer->addRow($rowD);
+$writer->addRow($rowE);
+$writer->addRow($rowF);
 
 $writer->close();
 ```
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -217,13 +217,14 @@ $writer = new Writer($options);
 $writer->openToFile('/tmp/file.xlsx');
 
 $levelOne = new RowAttributes(1);
-$rowA = Row::fromValues([1, 2, 3]);
-$options->setRowAttributes($rowA, $levelOne);
-$writer->addRow($rowA);
+$levelTwo = new RowAttributes(2);
 
-$rowB = Row::fromValues([3, 2, 1]);
-$options->setRowAttributes($rowB, $levelOne);
-$writer->addRow($rowB);
+$writer->addRow($options->setRowAttributes(Row::fromValues([1, 1, 1]), $levelOne));
+$writer->addRow($options->setRowAttributes(Row::fromValues([2, 2, 2]), $levelOne));
+$writer->addRow($options->setRowAttributes(Row::fromValues([3, 3, 3]), $levelTwo));
+$writer->addRow($options->setRowAttributes(Row::fromValues([4, 4, 4]), $levelTwo));
+$writer->addRow($options->setRowAttributes(Row::fromValues(), $levelOne));
+$writer->addRow(Row::fromValues());
 
 $writer->close();
 ```
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -196,6 +196,38 @@ $options->mergeCells(0, 1, 0, 2, $writer->getCurrentSheet()->getIndex());
 $writer->close();
 ```
 
+## Row Grouping/Outline
+
+Rows can be grouped with the XLSX writers:
+
+```php
+use OpenSpout\Common\Entity\Row;
+use OpenSpout\Writer\XLSX\Writer;
+use OpenSpout\Writer\XLSX\Options;
+use OpenSpout\Writer\XLSX\RowAttributes;
+
+$options = new Options();
+
+// Required or the generated file will be corrupted
+// Looks like empty rows/cell cannot have `defaultRowHeight` equal to 0
+// which will be added when the setting is not set.
+$options->DEFAULT_ROW_HEIGHT = 15;
+
+$writer = new Writer($options);
+$writer->openToFile('/tmp/file.xlsx');
+
+$levelOne = new RowAttributes(1);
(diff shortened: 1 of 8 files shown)


docs/documentation.md
```diff
@@ -181,6 +181,39 @@ $options->mergeCells(0, 1, 0, 2, $writer->getCurrentSheet()->getIndex());
...
+use OpenSpout\Writer\XLSX\Writer;
+use OpenSpout\Writer\XLSX\Options;
+use OpenSpout\Writer\XLSX\RowAttributes;
+
+$options = new Options();
+
+// Required or the generated file will be corrupted
+// Looks like empty rows/cell cannot have `defaultRowHeight` equal to 0
+// which will be added when the setting is not set.
+$options->DEFAULT_ROW_HEIGHT = 15;
```
Slamdunk, line 214: Can we something about this, and not require the user to specify it?

src/Writer/XLSX/Manager/WorksheetManager.php
```diff
@@ -142,6 +149,56 @@ private function addNonEmptyRow(Worksheet $worksheet, Row $row): void
...
+        $rowAttributes = ''
+            ." r=\"{$rowIndexOneBased}\""
+            .$this->getRowAttributes($row);
+
+        $rowXML = "<row{$rowAttributes}></row>";
+
+        $wasWriteSuccessful = fwrite($sheetFilePointer, $rowXML);
+        if (false === $wasWriteSuccessful) {
+            throw new IOException("Unable to write data in {$worksheet->getFilePath()}");
+        }
```
Slamdunk: I see you've just copy-pasted this and can't be tested: I say let's just assert it:
```suggestion
        assert(false !== $wasWriteSuccessful);
```

src/Writer/XLSX/Options.php
```diff
@@ -63,4 +70,22 @@ public function getMergeCells(): array
...
+        $this->rowsAttributes[$row] = $attributes;
+        $this->outlineMaxLevel = max($attributes->getOutlineLevel(), (int) $this->outlineMaxLevel);
+
+        return $row;
+    }
+
+    public function getRowAttributes(Row $row): ?RowAttributes
+    {
+        return $this->rowsAttributes[$row] ?? null;
+    }
```
Slamdunk: Nullable elements only leads to weak types; let's just return an empty `RowAttributes` and implement an `isEmpty` method on that object
```suggestion
    public function getRowAttributes(Row $row): RowAttributes
    {
        return $this->rowsAttributes[$row] ?? new RowAttributes();
    }
```

src/Writer/XLSX/RowAttributes.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\XLSX;
+
+use OpenSpout\Common\Exception\InvalidArgumentException;
+
+class RowAttributes
```
Slamdunk:
```suggestion
final class RowAttributes
```

src/Writer/XLSX/RowAttributes.php
```diff
@@ -0,0 +1,69 @@
...
+    private bool $visible;
+    private bool $collapsed;
+
+    /** @var null|int<0, 7> */
+    private ?int $outlineLevel;
+
+    /**
+     * @param null|int<0, 7> $outlineLevel
+     */
+    public function __construct(?int $outlineLevel, bool $collapsed = false, bool $visible = true)
```
Slamdunk: This whole class can just be a Value Object. Let's remove PHP `8.0.0` support and add an Enum for the OutlineLevel, then we can rewrite this class as:
```php
enum OutlineLevel: int
{
    case ONE    = 1;
    case TWO    = 2;
    case THREE  = 3;
    case FOUR   = 4;
    case FIVE   = 5;
    case SIX    = 6;
    case SEVEN  = 7;
}

final class RowAttributes
{
    public function __construct(
        public readonly ?OutlineLevel $OUTLINE_LEVEL = null,
        public readonly bool $COLLAPSED = false,
        public readonly bool $VISIBLE = true,
    ) {}
    
    public function isEmpty(): bool
    {
        return null === $this->OUTLINE_LEVEL
            && false === $this->COLLAPSED
            && true === $this->VISIBLE
        ;
    }
}
```

src/Writer/XLSX/Options.php
```diff
@@ -63,4 +70,22 @@ public function getMergeCells(): array
     {
         return $this->MERGE_CELLS;
     }
+
+    public function setRowAttributes(Row $row, RowAttributes $attributes): Row
```
Slamdunk: No fluent interfaces please
```suggestion
    public function setRowAttributes(Row $row, RowAttributes $attributes): void
```

Slamdunk (changes requested):
This looks awesome! Just few things to tweak and then :ship:

src/Writer/XLSX/RowAttributes.php
LastDragon-ru: > Let's remove PHP 8.0.0 support

No please, our project still on PHP 8.0)

docs/documentation.md
LastDragon-ru: The problem is in empty rows which are required to split outline regions (or Excell will merge them into one region) - with no height we will have `<sheetFormatPr defaultRowHeight="0"/>` and (web-based) Excel[^1][^2]  will report about cryptic error "not in a range" (or something like...) and file will not be opened at all. I've tried to add the cell, the cell with empty inline string, but no luck :(

Looks like height required in any case 🤷‍♂️

[^1]: standalone version seems can open it, but with "file was recovered" message
[^2]: Libreoffice Calc open file without any warnings

src/Writer/XLSX/RowAttributes.php
Slamdunk: https://github.com/openspout/openspout/blob/4.x/UPGRADE.md#upgrading-from-3x-to-40

I'm gonna do it anyway in 14 days https://www.php.net/supported-versions.php and I won't be able to switch `$OUTLINE_LEVEL` type in a minor release, so let's just create it as an Enum.

Until you are on 8.0 you can download your PR, modify it and apply it with https://github.com/cweagans/composer-patches to the latest `openspout:4.8` release

docs/documentation.md
Slamdunk: I don't like the idea that this library can produce a corrupted file if a particular option is not set.

Can you have a try to have a workaround within the code? I.E. if an emtpy row is added, the default row height gets set to a non zero value?

src/Writer/XLSX/RowAttributes.php
LastDragon-ru: Oh, I see... Unfortunately, I have only PHP 8.0 and cannot do it, sorry.

docs/documentation.md
LastDragon-ru: > `<row r="5" outlineLevel="1" customHeight="1" ht="15"></row>`

Doesn't work 😞 So I have no more ideas...

image

Slamdunk:
@LastDragon-ru I'm converting this to `Draft`: feel free to ping me when you figured out how to get it working

JonBaron78:
@LastDragon-ru How about a fallback to 15 like it is done in StyleManager::getContentXmlAutomaticStylesSectionContent() ?

`$defaultRowHeight = null === $this->options->DEFAULT_ROW_HEIGHT ? '15pt' : "{$this->options->DEFAULT_ROW_HEIGHT}pt";`

Maybe make fallback a constant?

jhogervorst:
@LastDragon-ru @Slamdunk I found this PR, as I'm looking into using Openspout in a project, but we need the "collapsed" functionality. It looks like this PR is quite old.

What is the status actually? Are the two currently open conversations the only remaining things before this PR can be merged? Or are there more issues?

Maybe we can help completing this PR 🙂

LastDragon-ru:
> What is the status actually?

@jhogervorst Unfortunately, I do not work on a project where this was used anymore, and have no time to update the PR. Sorry.

> Are the two currently open conversations the only remaining things before this PR can be merged?

As I understand/remember

1) Enum should be used instead `int` for level

2) Need some workaround to add empty row without `$options->DEFAULT_ROW_HEIGHT` setting by user, that will not crash ms excel and/or merge multiple regions.

Slamdunk:
> Maybe we can help completing this PR

Yes please: open a new one though.

jhogervorst:
> > Maybe we can help completing this PR
> 
> Yes please: open a new one though.

It turns out I made a mistake: for our project, we need to support collapsing _columns_, not _rows_.

So we'll not be completing this PR, but my colleague just opened a PR for collapsing columns: https://github.com/openspout/openspout/pull/275.

PR #115, opened 2022-11-24
add isEmpty function to Cell

steve3d:
in box\spout, there is a `isEmpty` function which is very convenient, but in openspout, the only way to test if a cell is empty is to use `instanceof EmptyCell`, but in order to test if is not I have to write `!($cell instanceof EmptyCell)` which is not a straight forward.

So I created this `isEmpty` function, to test if a cell is null or an empty string.

I don't know if using the `empty` function on EmptyCell is  appropriate. Please let me know if it's not.

anyway, thanks for this package.

### DIFF
--- a/src/Common/Entity/Cell.php
+++ b/src/Common/Entity/Cell.php
@@ -24,6 +24,8 @@ abstract class Cell
         $this->setStyle($style);
     }
 
+    abstract public function isEmpty(): bool;
+
     abstract public function getValue(): null|bool|string|int|float|DateTimeInterface|DateInterval;
 
     final public function setStyle(?Style $style): void
--- a/src/Common/Entity/Cell/BooleanCell.php
+++ b/src/Common/Entity/Cell/BooleanCell.php
@@ -21,4 +21,9 @@ final class BooleanCell extends Cell
     {
         return $this->value;
     }
+
+    public function isEmpty(): bool
+    {
+        return false;
+    }
 }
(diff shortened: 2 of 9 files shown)


Slamdunk:
> in order to test if is not I have to write `!($cell instanceof EmptyCell)` which is not a straight forward.

Thank you for your contribution but I think kind the opposite: type checks are much stronger and straightforward than APIs.

steve3d:
> Thank you for your contribution but I think kind the opposite: type checks are much stronger and straightforward than APIs.


I agree with this, But I have another question, in this repo, there is `StringCell`, `FormulaCell` and an special `EmptyCell`, so can I assume these is true?
- if a text cell is contains empty string, this cell will always be `EmptyCell`
- if a number cell contains empty string, this cell will aways be `EmptyCell`
- and all other special cell like `FormulaCell`, `DateTimeCell` contains empty string, then these cell must be `EmptyCell`?

what I want to make sure, whenever a cell contains empty content or `null` will always be `EmptyCell`?

If these is true, then there is no need for `isEmpty`, because in my project, null cells and empty cells must be considered empty no matter the cell type is numeric or date or something else.

Slamdunk:
See `Cell::fromValue` and its usages:

1. https://github.com/openspout/openspout/blob/4.x/src/Common/Entity/Cell.php#L39
2. For CSV: https://github.com/openspout/openspout/blob/4.x/src/Reader/CSV/RowIterator.php#L138
3. For ODS: https://github.com/openspout/openspout/blob/4.x/src/Reader/ODS/RowIterator.php#L318
4. For XLSX: https://github.com/openspout/openspout/blob/4.x/src/Reader/XLSX/RowIterator.php#L408

steve3d:
Thank you, this solved my question. then there is no need for this  `isEmpty`  function.

thanks again for this great package.

PR #117, opened 2022-12-15
Update docs howto stream directly without reading an excel source file

johan-hage:
I was missing in the docs howto stream on the fly without reading an already mentioned in the docs excel source file.
It's an easy example, but took me some time to figure out howto do this.


Slamdunk:
Thank you @johan-hage

PR #121, opened 2022-12-22
Set the SheetIterator index to zero on construct

Steart:
Because of the usage of type properties the `$currentSheetIndex` property must be set to zero by default.

Otherwise the following exception will be raised when calling the `current` method. `"Typed property OpenSpout\\Reader\\XLSX\\SheetIterator::$currentSheetIndex must not be accessed before initialization"`

### DIFF
--- a/src/Reader/XLSX/SheetIterator.php
+++ b/src/Reader/XLSX/SheetIterator.php
@@ -17,7 +17,7 @@ final class SheetIterator implements SheetIteratorInterface
     private array $sheets;
 
     /** @var int The index of the sheet being read (zero-based) */
-    private int $currentSheetIndex;
+    private int $currentSheetIndex = 0;
 
     /**
      * @param SheetManager $sheetManager Manages sheets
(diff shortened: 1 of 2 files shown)


Slamdunk:
The designed way to manually iterate over an `Iterator` requires to have `rewind($iterator);` as first call, witch already initializes it as expected

haroldiedema:
> The designed way to manually iterate over an `Iterator` requires to have `rewind($iterator);` as first call, witch already initializes it as expected

That's not true.

If a class property specifies a type and isn't explicitly set as a nullable value, the value MUST be initialized. There is no initialization value, therefore this is a bug.

This PR fixes that.

Besides, what's the problem with _not_ having to call rewind every time before you use this class? A class should always be in a valid state when it is constructed. This is broken in its current state.

Slamdunk:
> If a class property specifies a type and isn't explicitly set as a nullable value, the value MUST be initialized. There is no initialization value, therefore this is a bug.

The value is initialized, if you respect the `Iterator` API.

BTW I'll accept the PR as soon as a test is added.

Issue #122, opened 2022-12-25
Reading and writing cell-comments


jverelst:
Hi,

First of all, thanks for starting this fork and the fact that the code is still actively maintained! My question: are there any plans to implement cell comments (reading and writing) for XLSX files? I am working on a project that has heavy dependencies on PHPOffice and I am rewriting it to use spout for the obvious benefits of speed and memory usage. 

I did however notice that cell-comments are not supported, is that something that might be implemented in the future? I can also try to implement it myself and create a PR, but maybe somebody could give me some pointers on what the cleanest way would be to implement this?

Slamdunk:
Hi, my work here is to provide low level maintenance, I have no idea how comments are implemented in XLSX but I'll be very happy to accept PRs for them and to guide you to have the build green

jverelst:
@Slamdunk I've created a PR #123, so looking forward to your feedback

jverelst:
Since #123 has been merged, closing this issue.

PR #123, opened 2022-12-27
XLSX Writer: add Cell Comments support

jverelst:
As a follow-up to issue #122, I've tried to implement the writing-part of cell comments in this PR. Reading cell comments might be a bit more difficult since you don't want to load them into memory but instead read them streaming, but I am not sure that you can always do that, need to investigate that further.

Note that this is the first PR for me on this repo and I'm not too familiar with the strict typing from PHP8, so any feedback on code formatting/quality is appreciated and I will modify the PR if needed.

Some details on the implementation of comments:

### Files
Comments are split into files per sheet, and each sheet with id X requires 3 files to be created:
- `xl/commentsX.xml`
- `drawings/vmlDrawingX.vml`
- `xl/worksheets/_rels/sheetX.xml.rels`

The reason for this is that the comments are split into 2 parts: the actual text of the comment (in the XML)  and the way the panel is being rendered in Excel (this is the VML drawing). The rels file is required to tell the sheet where to find the drawing. 

### Implementation
First off, the model for a comment with it's related textruns were added in the `OpenSpout\Common\Entity\Comment` namespace:
- `Comment.php`
- `TextRun.php`

For actually writing these to files, I've chosen to create a new `CommentsManager.php` because that seemed to be the cleanest way to open/close and fill the 2 required files (XML + VML) while streaming the writes. Creating the 'rels' doesn't really belong here, because it is static and doesn't depend on the actual contents of the comments, so I've added that to the FilesystemHelper. 

This way, the number of places the current code is impacted is limited. All in all I'm quite pleased with how little code I had to modify :-)

Finally, I also updated the markdown documentation with both an example as well as some info on how to style comments.


Slamdunk:
Hi, I see that tests are missing: they are required for PR to be merged

jverelst:
I've reviewed the CI output and I've made 3 additional commits:
- Adding a unittest in the `WriterTest.php`
- Fixing all the output from phpstan
- Fixing all the output from php-cs-fixer

I saw that the CI run on the windows platform generated several warnings and a failure, which I am unsure about. The warnings are about the subdirectory `drawings` not being empty when a recursive delete is done, but I don't know why the file in that directory wasn't cleaned up. The method `Common/Helper/FileSystemHelper::deleteFolderRecursively` throws this warning but I can't see what is wrong with that, except that the order of files to delete might be wrong. That that would mean that the `RecursiveIteratorIterator` returns the files incorrectly on the windows platform and I'm not sure what could cause that. Do you have an idea?

Another failing test was about the number of rows written being incorrect, but that is something that I don't know how to fix as I am unsure whether my PR even touches that code. Any ideas?

Slamdunk:
Please run `make` to run the build locally, so you can locally inspect also Coding Standards and Static Analysis

jverelst:
> Please run `make` to run the build locally, so you can locally inspect also Coding Standards and Static Analysis

Ah, that helps. I've fixed and committed all the feedback that came out of this.

jverelst:
@Slamdunk still 2 errors left in the windows-latest run, that I don't know how to fix. Any ideas?

First one is about a lingering `drawings` directory (that I created in my PR), which isn't cleaned up corrrectly in the `deleteFolderRecursively` method. But I can't put my finger on the reason why. On my machine (OSX) and on the ubuntu CI run it works fine, on windows it fails. 

```
56) OpenSpout\Writer\XLSX\WriterWithStyleTest::testReuseBorders
rmdir(D:\a\openspout\openspout\tests\resources\generated_\temp\xlsx63ac4cd09b9450.90167794\xl\drawings): Directory not empty

D:\a\openspout\openspout\src\Common\Helper\FileSystemHelper.php:134
D:\a\openspout\openspout\src\Writer\XLSX\Helper\FileSystemHelper.php:104
D:\a\openspout\openspout\src\Writer\Common\Manager\AbstractWorkbookManager.php:185
D:\a\openspout\openspout\src\Writer\Common\Manager\AbstractWorkbookManager.php:146
D:\a\openspout\openspout\src\Writer\AbstractWriterMultiSheets.php:114
D:\a\openspout\openspout\src\Writer\AbstractWriter.php:147
D:\a\openspout\openspout\tests\Writer\XLSX\WriterWithStyleTest.php:586
D:\a\openspout\openspout\tests\Writer\XLSX\WriterWithStyleTest.php:535
```

Second one is even stranger, there is no reason that I can see to see why suddenly this rowcount isn't correct anymore. Also only happening in the windows build, not on OSX/linux. 

```
1) OpenSpout\Writer\XLSX\WriterTest::testShouldReturnWrittenRowCount
Failed asserting that 2 is identical to 5.

D:\a\openspout\openspout\tests\Writer\XLSX\WriterTest.php:607
```

src/Writer/XLSX/Manager/CommentsManager.php
```diff
@@ -114,6 +114,9 @@ public function closeWorksheetCommentFiles(Worksheet $sheet): void
 
         fwrite($commentFp, self::COMMENTS_XML_FILE_FOOTER);
         fwrite($drawingFp, self::DRAWINGS_VML_FILE_FOOTER);
+
+        fclose($commentFp);
```
jverelst, line 114: Oh, that was quite a stupid mistake :-(. Thanks for catching it 👍

Slamdunk:
Thank you @jverelst, I've move some code to be Value Objects with native types since we have them, but everything else it's ok!

:ship:

PR #124, opened 2022-12-28
XLSX Reader: add column width support

jverelst:
In this PR, I am adding support for reading the `<cols>` element that was recently added also to the XLSX Writer. It allows you to read the widths of columns, which I needed because I am trying to generate new XLSX files based on input files and require the column widths to remain the same.

A bit of example code on how this can be done:
```
$reader = new \OpenSpout\Reader\XLSX\Reader();
$reader->open('input.xlsx');

$options = new \OpenSpout\Writer\XLSX\Options();
$writer = new \OpenSpout\Writer\XLSX\Writer($options);
$writer->openToFile('output.xlsx');

  // Rewinding the iterator will read everything up until the first row, the <cols> element is placed before the sheetdata
  $sheet->getRowIterator()->rewind(); 

  // Read the column widths and copy them over to the new sheet
  foreach ($sheet->getColumnWidths() as $cw) {
    $options->setColumnWidthForRange($cw->width, $cw->start, $cw->end);
  }
 
  // Copy over all the cells
  foreach ($sheet->getRowIterator() as $row) {
      // do stuff with the row
      $cells = $row->getCells();
      $writer->addRow($row);
  }
}

$reader->close();
$writer->close();
```

To keep the code in sync with the way the columnwidths are written, I've added a new `Reader/Common/ColumnWidth` class, which is a direct copy of the `Writer/Common/ColumnWidth` class. Personally I think it would be better to remove this duplication, and move it to a generic class in `Common/Entity`. But I didn't want to start that refactorment without consultation. 

@Slamdunk if you think combining these two identical classes is better, then I will update my PR with that refactorment.


Slamdunk:
You should create dedicated branches for PR.
Please label the new commits into a branch, and rebase it onto the update upstream 4.x

PR #125, opened 2022-12-28
Adding column-widths to XLSX Reader

jverelst:
In this PR, I am adding support for reading the `<cols>` element that was recently added also to the XLSX Writer. It allows you to read the widths of columns, which I needed because I am trying to generate new XLSX files based on input files and require the column widths to remain the same.

A bit of example code on how this can be done:
```php
$reader = new \OpenSpout\Reader\XLSX\Reader();
$reader->open('input.xlsx');

$options = new \OpenSpout\Writer\XLSX\Options();
$writer = new \OpenSpout\Writer\XLSX\Writer($options);
$writer->openToFile('output.xlsx');

  // Rewinding the iterator will read everything up until the first row, the <cols> element is placed before the sheetdata
  $sheet->getRowIterator()->rewind(); 

  // Read the column widths and copy them over to the new sheet
  foreach ($sheet->getColumnWidths() as $cw) {
    $options->setColumnWidthForRange($cw->width, $cw->start, $cw->end);
  }
 
  // Copy over all the cells
  foreach ($sheet->getRowIterator() as $row) {
      // do stuff with the row
      $cells = $row->getCells();
      $writer->addRow($row);
  }
}

$reader->close();
$writer->close();
```

To keep the code in sync with the way the columnwidths are written, I've added a new `Reader/Common/ColumnWidth` class, which is a direct copy of the `Writer/Common/ColumnWidth` class. Personally I think it would be better to remove this duplication, and move it to a generic class in `Common/Entity`. But I didn't want to start that refactorment without consultation. 

@Slamdunk if you think combining these two identical classes is better, then I will update my PR with that refactorment.


src/Reader/XLSX/Sheet.php
```diff
@@ -47,6 +48,14 @@ public function getRowIterator(): RowIterator
         return $this->rowIterator;
     }
 
+    /**
+     * @return ColumnWidth[] a list of column-widths
+     */
+    public function getColumnWidths(): array
```
Slamdunk, line 58: Is this new API really needed? I mean, what's the use case of iterating column widths WITHOUT iterator rows first?

Slamdunk (changes requested):

src/Reader/XLSX/Sheet.php
jverelst: It's strictly not necessary, but it came from a more semantic point that I think the 'column widths' aspect is something that should is part of a sheet, not specifically of a rowiterator. 

In the current setup of the code, all the XML parsing of the sheet is done in the RowIterator, which makes sense if you are only concerned with the row data. But the sheet-metadata (such as column widths) is now also being parsed by the RowIterator, ideally I think that should happen somewhere else: on sheet level. Because now you first have to `rewind()` the RowIterator to make it read the header of the sheet, to be able to retrieve the column widths. 

That's the reason I put the PR back into draft, because I'm not too happy with this, but refactoring the XML parsing into 2 parts (one bit for reading the header of the sheet and one for iterating over the rows) seemed like quite a bit of work to achieve something quite simple as getting the column sizes. 

So by introducing the method on Sheet, it should be possible to do this refactorment later, without breaking the API. That also means that the `RowIterator::getColumWidths()` should be marked `@internal` I guess.

jverelst:
@Slamdunk I've reworked the PR, adding a new `SheetHeaderReader` that reads the header of the XLSX sheet. I think that's cleaner, because reading sheet metadata in the RowIterator just didn't feel like the right place. This way, reading sheet metadata can in the future easily be extended. 

I've kept the `getColumnWidths()` on the `Sheet` class, because I didn't want to expose the `SheetHeaderReader`

jverelst:
@Slamdunk do you have any comments/suggestions for this PR, or could it be merged?

Slamdunk:
I've merged in the upstream branch an edit that runs Infection only on diff, so its feedback is now much more valuable.

Can you rebase your branch on the updated `4.x` please?

src/Reader/XLSX/SheetHeaderReader.php
```diff
@@ -0,0 +1,119 @@
...
+        XMLProcessor $xmlProcessor
+    ) {
+        $this->filePath = $filePath;
+        $this->sheetDataXMLFilePath = $this->normalizeSheetDataXMLFilePath($sheetDataXMLFilePath);
+        $this->xmlReader = $xmlReader;
+
+        // Register all callbacks to process different nodes when reading the XML file
+        $this->xmlProcessor = $xmlProcessor;
+        $this->xmlProcessor->registerCallback(self::XML_NODE_COL, XMLProcessor::NODE_TYPE_START, [$this, 'processColStartingNode']);
+        $this->xmlProcessor->registerCallback(self::XML_NODE_SHEETDATA, XMLProcessor::NODE_TYPE_START, [$this, 'processSheetDataStartingNode']);
```
Slamdunk, line 54: Infection says that removing this line doesn't result in any test failing.
I'm fine with the other 3 Infection reports not being addressed, but this one puzzles me.

If no test fails, what's the purpose of this callback?

Slamdunk (changes requested):

src/Reader/XLSX/SheetHeaderReader.php
jverelst: The reason I added it is because of performance. Once you get to the `<sheetData>` element, the header reading should stop: the row data starts. If you have a very large sheet with a lot of rows, it makes no sense to read all those rows when there is no header information left.

So on lines 63/64, I do:
```
        // Now read the entire header of the sheet, until we reach the <sheetData> element
        $this->xmlProcessor->readUntilStopped();
```

By setting this callback, the header processing stops at the `<sheetData>` element, and you can close the reader. 

I don't know how to write a test for that though.

Slamdunk:
Thank you @jverelst

PR #126, opened 2022-12-30
XLSX Comments: fixing typo for `italic` tag

jverelst:
During implementation of the Comments writing in my own project, I noticed a typo in the code: italic textruns should be identified by `<i/>` and not `<b/>`


Slamdunk:
Hi, can you add a test please, so we can be sure the bug will not be reintroduced in future?

jverelst:
> Hi, can you add a test please, so we can be sure the bug will not be reintroduced in future?

Good point, test added.

PR #134, opened 2023-01-19
CSV: write `false` as `0`

### LINKED ISSUE #133: Inconsistent boolean values when writing to CSV
When a `BooleanCell` containing `false` is written to a CSV file the resulting value will be empty (`''` to be exact), whereas a `BooleanCell` containing `true` will result in `'1'`. This is because all [cell values are cast to strings here](https://github.com/openspout/openspout/blob/4.x/src/Writer/CSV/Writer.php#L53) and PHP casts `false` to an empty string ([reference](https://www.php.net/manual/en/language.types.string.php#language.types.string.casting)). I would expect `false` to result in `'0'` in the CSV.

Now this could be solved in consumer code by explicitly casting booleans to integers, but ideally we'd fix this in the CSV writer. I couldn't find any tests covering this behaviour, so I'm unsure whether it's intentional or accidental.

What are your thoughts? Could this be a bug fix?

devfrey:
Fixes #133

### DIFF
--- a/src/Writer/CSV/Writer.php
+++ b/src/Writer/CSV/Writer.php
@@ -50,13 +50,11 @@ final class Writer extends AbstractWriter
     protected function addRowToWriter(Row $row): void
     {
         $cells = array_map(static function (Cell\BooleanCell|Cell\EmptyCell|Cell\NumericCell|Cell\StringCell|Cell\FormulaCell $value): string {
-            $value = $value->getValue();
-
-            if (false === $value) {
-                return '0';
+            if ($value instanceof Cell\BooleanCell) {
+                return (string) (int) $value->getValue();
             }
 
-            return (string) $value;
+            return (string) $value->getValue();
         }, $row->getCells());
 
         $wasWriteSuccessful = fputcsv(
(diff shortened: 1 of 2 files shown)


src/Writer/CSV/Writer.php
```diff
@@ -50,7 +50,13 @@ protected function openWriter(): void
...
     {
         $cells = array_map(static function (Cell\BooleanCell|Cell\EmptyCell|Cell\NumericCell|Cell\StringCell|Cell\FormulaCell $value): string {
-            return (string) $value->getValue();
+            $value = $value->getValue();
+
+            if (false === $value) {
+                return '0';
+            }
+
+            return (string) $value;
```
Slamdunk: I'd rather check for types instead of values:
```suggestion
            if ($value instanceof Cell\BooleanCell) {
                return (string) (int) $value->getValue();
            }

            return (string) $value->getValue();
```

Slamdunk (changes requested):

Slamdunk (approved):

Slamdunk:
Thank you @devfrey

PR #140, opened 2023-03-09
Added support for setting column widths per sheet

### LINKED ISSUE #88: How to set column width for certain sheet only
Hello,

Is there any options to set column width for certain sheet only ?

mlntn:
For issue #88

### DIFF
--- a/src/Writer/Common/Entity/Sheet.php
+++ b/src/Writer/Common/Entity/Sheet.php
@@ -168,7 +168,7 @@ final class Sheet
     /**
      * @param positive-int ...$columns One or more columns with this width
      */
-    final public function setColumnWidth(float $width, int ...$columns): void
+    public function setColumnWidth(float $width, int ...$columns): void
     {
         // Gather sequences
         $sequence = [];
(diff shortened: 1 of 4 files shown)


src/Writer/Common/Entity/Sheet.php
```diff
@@ -160,4 +164,45 @@ public function getAutoFilter(): ?AutoFilter
     {
         return $this->autoFilter;
     }
+
+    /**
+     * @param positive-int ...$columns One or more columns with this width
+     */
+    final public function setColumnWidth(float $width, int ...$columns): void
```
Slamdunk: Overall this PR looks good, but I don't like the complexity under the hood of this function signature.
I'd prefer force the user call this multiple times more than force ourself to maintain such a complex function content.
```suggestion
    final public function setColumnWidth(float $width, int $column): void
```

Slamdunk (changes requested):
:muscle:

src/Writer/Common/Entity/Sheet.php
mlntn: Thanks for reviewing!

This decision was based on an existing method: https://github.com/openspout/openspout/blob/90ceeefef3750dd2afab95689ed2107dd576905e/src/Writer/Common/AbstractOptions.php#L30 Do you want me to deviate from that?

Slamdunk: Didn't remember that (aweful) code, that's for pointing it out

Slamdunk (changes requested):
Please run CS Fix and we are done

Slamdunk (approved):

PR #143, opened 2023-03-30
fix "Undefined array key 1"

### LINKED ISSUE #141: Error "Undefined array key 1" when reading some xlsx
https://github.com/openspout/openspout/blob/e01cab951c04ef184c567a880290997e94386f94/src/Reader/XLSX/RowIterator.php#L299-L303
Some xlsx files can be read normally by Office Excel, but openspout will return an empty string '' here $spans, causing it to pass the if judgment and then report an error “Undefined array key 1”.

https://github.com/box/spout/blob/84596668410bea89d21aa9867b91e1550e359329/src/Spout/Reader/XLSX/RowIterator.php#L258-L262

If the judgment method of the original version of Spout is modified, the file can be read normally.

zxl409774878:
Some xlsx files will get the empty string $spans to raise this error, add the empty string judgment to improve compatibility.


Slamdunk:
Thank you @zxl409774878

PR #145, opened 2023-04-13
Fixed dynamic properties deprecation in PHP 8.2

meirzamoodle:
Scanning OpenSpout files using PHPStan and found there are some properties that need to added to comply with PHP 8.2

### DIFF
--- a/src/Reader/Wrapper/XMLReader.php
+++ b/src/Reader/Wrapper/XMLReader.php
@@ -14,6 +14,15 @@ final class XMLReader extends \XMLReader
     use XMLInternalErrorsHelper;
 
     public const ZIP_WRAPPER = 'zip://';
+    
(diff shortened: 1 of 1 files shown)


Slamdunk:
Mmm, those properties are already there:
1. https://www.php.net/manual/en/class.xmlreader.php#xmlreader.props.nodetype
2. https://github.com/php/php-src/blob/702b0cc7a428b628ccb25fd4c8271b2b3eb4e242/ext/xmlreader/php_xmlreader.stub.php#L144
3. https://github.com/php/php-src/blob/702b0cc7a428b628ccb25fd4c8271b2b3eb4e242/ext/dom/php_dom_arginfo.h#L1050-L1054

This is a PHPStan bug already reported: https://github.com/phpstan/phpstan/issues/8629

PR #156, opened 2023-05-22
XLSX: Use platform appropriate directory separator for shared strings temp filename

### LINKED ISSUE #155: Slow FileBasedStrategy on Windows
[This condition](https://github.com/openspout/openspout/blob/8f0e9576bdccd6f512fa614216fd58db71337992/src/Reader/XLSX/Manager/SharedStringsCaching/FileBasedStrategy.php#L124) seems to be always passing on Windows, because `$inMemoryTempFilePath` is normalized using `realpath()` while the `$tempFilePath` is not.

| variable | value |
|---|---|
| `$this->inMemoryTempFilePath` | `C:\...\Temp\sharedstrings64676f752d764\sharedstrings0` |
| `$tempFilePath` | `C:\...\Temp\sharedstrings64676f752d764/sharedstrings0` |

Notice the incorrect directory separator (a slash rather than a backslash) ahead of the base name part.

This causes the temporary file to be re-loaded on every single shared string access, making it very slow, especially on XLSX's with a lot of shared strings.

adawolfa:
fixes #155

### DIFF
--- a/src/Reader/XLSX/Manager/SharedStringsCaching/FileBasedStrategy.php
+++ b/src/Reader/XLSX/Manager/SharedStringsCaching/FileBasedStrategy.php
@@ -164,7 +164,7 @@ final class FileBasedStrategy implements CachingStrategyInterface
     {
         $numTempFile = (int) ($sharedStringIndex / $this->maxNumStringsPerTempFile);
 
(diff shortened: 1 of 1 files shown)


Slamdunk:
Thank you

PR #160, opened 2023-06-07
Fix the suggested PHP extensions

stof:
`mb_convert_encoding` is part of ext-mbstring, not of ext-intl.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,8 +47,8 @@
         "phpunit/phpunit": "^10.1.3"
     },
     "suggest": {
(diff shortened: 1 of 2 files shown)


Slamdunk (changes requested):
You are right, thank you.

Can you also fix the `intl` reference in https://github.com/openspout/openspout/blob/e4179a2b68ccd35fad52e60b0c7ebbf8ad9fe677/src/Common/Helper/EncodingHelper.php#L186 please?

stof:
@Slamdunk done

Slamdunk (approved):

Slamdunk:
Thank you :muscle:

PR #162, opened 2023-06-19
Implemented CachingStrategyFactoryInterface

### LINKED ISSUE #161: Prevent xlsx shared strings file based caching
I'm currently updating my spout implementation from version 3 to 4 and I stumbled about a undocumented (and maybe unintended) breaking change. 

For XLSX Shared Strings you have the class OpenSpout\Reader\XLSX\Manager\SharedStringsCaching\CachingStrategyFactory. This class trys to decide whether to use the InMemoryStrategy or the FileBasedStrategy. Previously this factory class was not final and extendable. We used this to prevent the usage of the FileBasedStrategy because it slows down to whole process extremly - we would rather let the script die instead of slowing it down. So previously we just extended the CachingStrategyFactory with our own class:

```php
class CachingStrategyFactory extends \OpenSpout\Reader\XLSX\Manager\SharedStringsCaching\CachingStrategyFactory
{
    /**
     * Returns the best shared strings caching strategy.
     *
     * @param int|null $sharedStringsUniqueCount Number of unique shared strings (NULL if unknown)
     * @param string $tempFolder Temporary folder where the temporary files to store shared strings will be stored
     * @param HelperFactory $helperFactory Factory to create helpers
     * @return CachingStrategyInterface The best caching strategy
     */
    public function createBestCachingStrategy($sharedStringsUniqueCount, $tempFolder, $helperFactory)
    {
        return new InMemoryStrategy($sharedStringsUniqueCount);
    }
}
```

In version 4 the CachingStrategyFactory is now final and don't has any usable interfaces which could be used instead. The only way I could hack the old behaviour together would be to create a CachingStrategyFactory with a fake memory limit - something like this:

```php
$cachingStrategyFactory = new CachingStrategyFactory(new MemoryLimit('1tb'));
$reader =  new XlsxReader(null, $cachingStrategyFactory);
```

I currently see three options to resolve this:

1. Remove the final from the CachingStrategyFactory class again.
2. Create a CachingStrategyFactoryInterface and replace the typehints for CachingStrategyFactory with the interface. This shouldn't break anything. 
3. Add an option to the CachingStrategyFactory to enforce any CachingStrategy.

Option 2 would probably the cleanest solution. Option 1 is by far the easiest and fastest.

MaximilianKresse:
See #161

### DIFF
--- a/src/Reader/XLSX/Manager/SharedStringsCaching/CachingStrategyFactory.php
+++ b/src/Reader/XLSX/Manager/SharedStringsCaching/CachingStrategyFactory.php
@@ -7,7 +7,7 @@ namespace OpenSpout\Reader\XLSX\Manager\SharedStringsCaching;
 /**
  * @internal
  */
-final class CachingStrategyFactory
+final class CachingStrategyFactory implements CachingStrategyFactoryInterface
 {
     /**
      * The memory amount needed to store a string was obtained empirically from this data:.
--- a/src/Reader/XLSX/Manager/SharedStringsCaching/CachingStrategyFactoryInterface.php
+++ b/src/Reader/XLSX/Manager/SharedStringsCaching/CachingStrategyFactoryInterface.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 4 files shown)


Slamdunk:
Thank you :muscle:

PR #101, opened 2022-10-18
Add PHP 8.2 support

acrobat:
Allow php 8.2 and add it to the test matrix. This PR fixes also a deprecation in the custom stream wrapper.

I'm not sure what causes the other failing tests as I can't reproduce them on my local php 8.2 installation 🤔

Issue #110, opened 2022-11-09
Create release of v3 for PHP 8.2


freekmurze:
Could you create a new release of v3 that allow the use of PHP 8.2
It's highly likely that v3 will just work on PHP 8.2

That would fix the tests of my dependant package: https://github.com/spatie/simple-excel

Slamdunk:
I'm sorry `v3` has been EOLed.

Upgrade to `v4` should be easy tough: https://github.com/openspout/openspout/blob/4.x/UPGRADE.md

cmanley:
It would be nice to have a little PHP version support overlap next time.
Why: because currently as it stands one currently can't gracefully upgrade from Debian 11 (PHP 7.4) to 12 (PHP 8) without simultaneously having to upgrade and release all code that uses OpenSpout.
If one uses Docker then this process is less of an issue obviously.

PR #164, opened 2023-06-25
Add PHP 8.2 support to 3.x branch

iansltx:
This is a re-roll of https://github.com/openspout/openspout/pull/101 against the 3.x branch, allowing users to have a version that supports both PHP 7.4 (the last version compatible with the old Box-maintained lib) and 8.2, vs. needing an intermediate runtime version step (7.4 -> 8.1, then lib version 3.x -> 4.x, then 8.1 -> 8.2).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
       matrix:
         php-version:
           - "8.1"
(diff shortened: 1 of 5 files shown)


Slamdunk:
Duplicate of https://github.com/openspout/openspout/issues/110

PR #165, opened 2023-07-12
Changed worksheet filepath in xls/xl/worksheets

urlportal:
Hello. I needed to rename the prefix DEFAULT_SHEET_NAME_PREFIX of the sheet name in the file src/Writer/Common/Entity/Sheet.php. I redefined this file, changed the prefix. The Excel file has been created successfully. It had the name of the sheet that I needed. But the data on the sheet was not displayed. I studied the xlsx files created by Microsoft Excel and found out that in the xls/xl/worksheets directory, the sheets are always named sheet1.xml , sheet2.xml regardless of how the sheets are named.
Also in your file src/Writer/XLSX/Helper/FileSystemHelper.php on line 193: "worksheets/sheet'.$worksheetId.'.xml", i.e. you are correctly linking to an XML file with the name sheet1.xml , sheet2.xml... The filename is hardcoded here. I suggest that in the correction, specify a hard name. This will allow you to change the name of the sheet.
I do not know if you still support the 3.x branch, but I think my fix would not hurt.
Thanks for your work

### DIFF
--- a/src/Writer/XLSX/Manager/WorkbookManager.php
+++ b/src/Writer/XLSX/Manager/WorkbookManager.php
@@ -35,7 +35,7 @@ class WorkbookManager extends WorkbookManagerAbstract
     {
         $worksheetFilesFolder = $this->fileSystemHelper->getXlWorksheetsFolder();
 
-        return $worksheetFilesFolder.'/'.strtolower($sheet->getName()).'.xml';
+        return $worksheetFilesFolder.'/sheet' . ($sheet->getIndex() + 1) . '.xml';
     }
 
     /**


Slamdunk:
Thank you for your contribution but indeed `3.x` branch is not supported anymore.

May you target your PR against `4.x`, if the patch is applicable?

urlportal:
In our project, we are using PHP 7.4. I think many people are still using PHP 7.4. I was surprised that it is not possible to rename the sheet. I was considering starting using your library, but version 4.X only works with PHP ^8.1.
> May you target your PR against `4.x`, if the patch is applicable?

I can't consider preparing a fix for the 4.X branch, because there are a lot of changes in the branch 4.X, and I don't have projects with PHP 8.1 yet. Maybe I'll do it later

Slamdunk:
> but version 4.X only works with PHP ^8.1

Version 4 requires PHP ^8.0:
https://github.com/openspout/openspout/blob/1f5d8500f3fdcc023543293f052a8beb4a92de78/composer.json#L31

Slamdunk:
Branch `4.x` is already fixed: https://github.com/openspout/openspout/blob/ef5396bfd44b3dfa092e78458966d25c7a2c931a/src/Writer/Common/Manager/AbstractWorkbookManager.php#L176

And since branch `3.x` has been EOLed, I'm not merging this.

I you really need your patch, have a look at [`cweagans/composer-patches`](https://github.com/cweagans/composer-patches)

PR #166, opened 2023-08-01
Fix date pattern detection/double-quoted strings

MalteWunsch:
Excel uses double-quoted strings to indicate raw text. So when looking for date format patterns in custom cell formats, we should ignore double-quoted strings to prevent false-positives.


MalteWunsch:
@Slamdunk The php-cs-fixer fails are unrelated - should I include their fixes anyway?

Slamdunk:
Yes please

MalteWunsch:
Included ✅

Slamdunk:
Can you give me an Excel with such data, so I can do a manual test before merging please?

MalteWunsch:
@Slamdunk Here you are: [test.xlsx](https://github.com/openspout/openspout/files/12317223/test.xlsx)

PR #169, opened 2023-08-21
Reader\XLSX: retrieve formulas as-is, but keep computed value available when possible

### LINKED ISSUE #168: How to get raw formula from cell?
Hey there.

Thank you for making this. It's helped me out a number of times so far.

**How do I get the raw value/formula of a cell?** I'm using references in my document (e.g. `='Sheet name'!$Row$Cell`), but when I call `->getValue()` I get the dereferenced value from the sheet. Is there a way to get the raw formula?

Slamdunk:
Fix https://github.com/openspout/openspout/issues/168

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,24 +23,22 @@ csfix: vendor
 static-analysis: vendor
 	php -d zend.assertions=1 vendor/bin/phpstan analyse $(PHPSTAN_ARGS)
 
-coverage/junit.xml: vendor $(SRCS) Makefile phpunit.xml
+coverage/ok: vendor $(SRCS) Makefile phpunit.xml
 	chmod -fR u+rwX tests/resources/generated_* || true
 	rm -fr tests/resources/generated_*
-	php \
+	(php \
 		-d zend.assertions=1 \
 		-d open_basedir="$(realpath .)" \
 		-d sys_temp_dir="$(realpath .)" \
 		vendor/bin/phpunit \
-		--coverage-xml=coverage/coverage-xml \
-		--coverage-html=coverage/html \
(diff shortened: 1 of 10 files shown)


mathielen:
Hi there,

Not sure if you have noticed but this breaks backwards compatibility. When just using `getValue` on the cell without knowing what the cell's implementation is (here: a `FormulaCell`) it will now return the formula instead of the computed value. My code still expects to receive the computed value here.

I suggest to keep `getValue` returning the computed value and replace `getComputedValue` to `getFormulaValue` and return the formula only when *this* method is invoked.

Best regards
Markus

Slamdunk:
> Not sure if you have noticed but this breaks backwards compatibility. When just using `getValue` on the cell without knowing what the cell's implementation is (here: a `FormulaCell`) it will now return the formula instead of the computed value.

This is expected: the old behaviour was a bug. Idempotency is important: if I read and re-write the file, I should get the exact same file back; this PR gets this goal right.

> My code still expects to receive the computed value here.

The computed value may or may not be there, it is not guaranteed. Checking for `FormulaCell` type and `getComputedValue()` return value to be not-null is expected to be required for a reliable usage.

mathielen:
Fair enough. That also makes sense. I will adapt.

Thank you!

dakira:
@Slamdunk Hi. I just ran into this and all my projects which I updated broke down when updating from openspout v4.15.1 onward.

If existing behavior is changed, than it is a backwards compatibility breaking change, even if the existing behavior is buggy.

I appreciate your reasoning and agree with it, but it *would* be nice to put such a change in a major version so that other peoples software that uses this doesn't break. 😘

PR #171, opened 2023-08-22
XLSX: Add page margin, page orientation and paper size options

cresjie:


cresjie:
update code for php-cs-fixer

cresjie:
added unit test

cresjie:
update code for php-cs-fixer and infection

src/Writer/XLSX/Options.php
```diff
@@ -63,4 +70,103 @@ public function getMergeCells(): array
...
+     * @param float $header inches
+     * @param float $footer inches
+     */
+    public function setPageMargins(
+        float $top = 0.75,
+        float $right = 0.7,
+        float $bottom = 0.75,
+        float $left = 0.7,
+        float $header = 0.3,
+        float $footer = 0.3
```
Slamdunk: Let's use a dedicated object like the [`MergeCell`](https://github.com/openspout/openspout/blob/v4.16.0/src/Writer/XLSX/MergeCell.php) one

src/Writer/XLSX/Options.php
```diff
@@ -63,4 +70,103 @@ public function getMergeCells(): array
...
+     */
+    public function getPageMargins(): array
+    {
+        return $this->pageMargins;
+    }
+
+    /**
+     * @param string $orientation portrait|landscape
+     */
+    public function setPageOrientation(string $orientation): void
```
Slamdunk: Replace the `string` with a dedicated `Enum`

src/Writer/XLSX/Options.php
```diff
@@ -63,4 +70,103 @@ public function getMergeCells(): array
...
+    {
+        $this->pageSetup['orientation'] = $orientation;
+    }
+
+    public function setPaperSize(string $size): void
```
Slamdunk: Replace the `string` with a dedicated `Enum`

Slamdunk (changes requested):
Changes look very promising, let's try to add more precise types please :muscle:

cresjie:
updated to use enum and dedicated class for PageMargin.

by the way, any idea why running this command locally:
```
git fetch origin --no-tags --prune --depth=1 4.x 3f0c3dbbfe2db2f6215eafddd9614b693ba1046b
make code-coverage INFECTION_ARGS='--logger-github'
```



outputs:
```
0 mutations were generated:
       0 mutants were killed
       0 mutants were configured to be ignored
       0 mutants were not covered by tests
       0 covered mutants were not detected
       0 errors were encountered
       0 syntax errors were encountered
       0 time outs were encountered
       0 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 0%
         Mutation Code Coverage: 0%
         Covered Code MSI: 0%
```

cresjie:
update code for infection MSI

Slamdunk:
@cresjie whenever you push a commit I get notified, no need to comment every commit to ping me 😉

src/Writer/XLSX/Options.php
Slamdunk: We can directly require a `PageMargin` object as the argument

src/Writer/XLSX/Options.php
```diff
@@ -63,4 +71,65 @@ public function getMergeCells(): array
...
+            $top,
+            $right,
+            $bottom,
+            $left,
+            $header,
+            $footer,
+        );
+    }
+
+    public function getPageMargin(): PageMargin
```
Slamdunk: Returning `?PageMargin` seems enough for me to the rid of `hasPageMargin` method

src/Writer/XLSX/Options.php
```diff
@@ -63,4 +71,65 @@ public function getMergeCells(): array
...
+    {
+        return $this->pageMargin;
+    }
+
+    public function hasPageMargin(): bool
+    {
+        return (bool) $this->pageMargin;
+    }
+
+    public function setPageOrientation(EnumPageOrientation $orientation): void
```
Slamdunk: Same as `PageMargin` we can introduce a `PageSetup` class that accepts both orientation and papere size, and a unique `setPageSetup` method, so we can get rid of the array.

src/Writer/XLSX/Options/PageMargin.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\XLSX\Options;
+
+/**
+ * @internal
+ */
```
Slamdunk:
```suggestion
```
This cannot be internal

Slamdunk (changes requested):

src/Writer/XLSX/Options.php
cresjie: yup, i agree with that, but the phpstan will throw an error.
```
Only booleans are allowed in an if condition, OpenSpout\Writer\XLSX\Options\PageMargin given
```

any advice with that error?

Slamdunk:
```php
$pageMargin = $options->getPageMargin();
if (null !== $pageMargin) {
    // ...
}
```

cresjie: i think i'll just use isset()

Slamdunk: No please, `null !==` is strictier than `isset`

Slamdunk (approved):

PR #173, opened 2023-09-11
Reader\XLSX: Format formulas computed value

roberts-santars:
Fixed issue with XLSX Formulas `computedValue` as introduced in https://github.com/openspout/openspout/pull/169
**Before** changes in https://github.com/openspout/openspout/pull/169:
- `FormulaCell::getValue()` returns formatted value of cell formula result

**After** changes in https://github.com/openspout/openspout/pull/169:
- `FormulaCell::getValue()` replacement `FormulaCell::getComputedValue()` ignores cell formatting as was returned previously by `FormulaCell::getValue()`

Adjusted that `FormulaCell::getComputedValue()` can return `DateTimeImmutable|float|int|string|null` according to cell formatting, similarly as value was formatted before.


Slamdunk:
Thank you @roberts-santars

PR #177, opened 2023-09-15
Writer\XLSX: add support for DateInterval

### LINKED ISSUE #176: Excel support for DateInterval
The `OpenSpout\Writer\XLSX\Manager\WorksheetManager::getCellXML` method doesn't handle the `DateIntervalCell`, leaving the XML with broken unclosed tags:

image

uncaught:
Adds Excel support for DateIntervals (fixes #176).

I'm not sure whether there is an option to specify formats for Excel only. Otherwise I would have requested a valid Excel-Format in the DateIntervalCell constructor.

Without the right format on the Style, it won't work in Excel.

### DIFF
--- a/src/Common/Entity/Cell/DateIntervalCell.php
+++ b/src/Common/Entity/Cell/DateIntervalCell.php
@@ -12,6 +12,12 @@ final class DateIntervalCell extends Cell
 {
     private DateInterval $value;
 
+    /**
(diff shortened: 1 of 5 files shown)


uncaught:
Sorry for the many pushes - I don't have the tools setup right for this project in my WSL/docker.

Slamdunk:
Ok, please just don't force push, regular commits one over another helps me with the review process

PR #185, opened 2023-09-25
Add readonly property to prevent modification

peter279k:
# Changed log

- According to the [reference](https://wiki.php.net/rfc/readonly_properties_v2), it should add `readonly` property to prevent modification after initializing the property.


Slamdunk (approved):

Slamdunk:
Thank you again :muscle:

PR #187, opened 2023-09-26
Apply null to be strict string

peter279k:
# Changed log

- Apply the null to be strict string because it can avoid presenting deprecated message and pass null parameter to these functions.


Slamdunk (changes requested):
I don't think we need this fix

peter279k:
You're right. I think these parameters with casting string will not present deprecated warning messages after looking at these codes.

PR #188, opened 2023-10-04
Reader\XLSX: add support for DateInterval

### LINKED ISSUE #197: Add support for DateInterval in XLSX Reader
The XLSX writer got support for DateIntervals with #177, the reader is still missing the respective support though.

I've added this with #188.

uncaught:
- Adds support to read DateInterval from Excel files (closes #197)
- Detection of DateInterval is based on custom date formats beginning with a bracket hour/minute/second format, so these are all the possible formats that would produce a DateInterval instead of a DateTimeImmutable now:
    - `[hh]`, `[hh]:mm`, `[hh]:mm:ss`
    - `[h]`, `[h]:mm`, `[h]:mm:ss`
    - `[mm]`, `[mm]:ss`, 
    - `[m]`, `[m]:ss`
    - `[ss]`, `[s]`
- Changed format of cell `I3` in `tests/resources/xlsx/sheet_with_same_numeric_value_date_formatted_differently.xlsx` from `[h]:mm:ss` to `h:mm:ss` so that it is not detected as DateInterval.
- Added new tests for all possible DateInterval formats:\
image

Although I doubt that anyone uses brackets format for a regular date, this PR should probably be considered a breaking change because it changes the outcome of existing reads in case such formats were used.

-----

- Also adds support for negative DateIntervals in Writer\XLSX

-----

Unrelated changes to fix phpunit tests:

- I had to update the `StyleManager::getNumberFormatCode()` to not return `null` (this breaks with PHP 8.2)
- I had to modify `testReadShouldBeProtectedAgainstBillionLaughsAttack` because its expected memory limit was exceeded already prior to reading the ods file. I changed it so that the expected memory peak is less than 2 MB higher than the peak prior to reading. I hope this suffices for the test purpose.


src/Reader/XLSX/Helper/DateIntervalFormatHelper.php
```diff
@@ -0,0 +1,96 @@
...
+declare(strict_types=1);
+
+namespace OpenSpout\Reader\XLSX\Helper;
+
+use DateInterval;
+
+final class DateIntervalFormatHelper
+{
+    /**
+     * See https://www.php.net/manual/de/dateinterval.format.php.
```
Slamdunk:
```suggestion
     * @see https://www.php.net/manual/en/dateinterval.format.php.
```

src/Reader/XLSX/Helper/DateIntervalFormatHelper.php
```diff
@@ -0,0 +1,96 @@
...
+            $interval->invert = 1;
+        }
+
+        return $interval;
+    }
+
+    public static function isDurationFormat(string $excelFormat): bool
+    {
+        // Only consider formats with leading brackets as valid duration formats (e.g. "[hh]:mm", "[mm]:ss", etc.):
+        return (bool) preg_match('/^(\[hh?](:mm(:ss)?)?|\[mm?](:ss)?|\[ss?])$/', $excelFormat);
```
Slamdunk:
```suggestion
        return 1 === preg_match('/^(\[hh?](:mm(:ss)?)?|\[mm?](:ss)?|\[ss?])$/', $excelFormat);
```

src/Writer/XLSX/Helper/DateIntervalHelper.php
```diff
@@ -25,11 +25,28 @@ public static function toExcel(DateInterval $interval): float
...
+        if (1 === $interval->invert) {
+            self::negate($days);
+        }
+
+        return $days;
+    }
+
+    /**
+     * The infection rule MulEqual doesn't understand that multiplying or dividing by -1 is mathematically identical.
+     * See https://github.com/infection/infection/issues/1884.
```
Slamdunk: Now that your PR has been merged and released, can we get back to `$value *= -1;` inside the `toExcel()` method, removing the additional `negate()`?

tests/Writer/XLSX/Helper/DateIntervalHelperTest.php
```diff
@@ -41,6 +41,14 @@ public static function provideDateIntervalToExcelCases(): array
             [DateInterval::createFromDateString('1 second'), 1 / 24 / 60 / 60],
             [DateInterval::createFromDateString('2 seconds'), 2 / 24 / 60 / 60],
             [new DateInterval('PT12H30M0S'), 0.5 + 30 / 24 / 60],
+            [self::invert(new DateInterval('PT12H0M0S')), -0.5],
```
Slamdunk, line 44:
```suggestion
            [$this->invert(new DateInterval('PT12H0M0S')), -0.5],
```

tests/Writer/XLSX/Helper/DateIntervalHelperTest.php
```diff
@@ -41,6 +41,14 @@ public static function provideDateIntervalToExcelCases(): array
...
         ];
     }
+
+    private static function invert(DateInterval $interval): DateInterval
```
Slamdunk, line 48:
```suggestion
    private function invert(DateInterval $interval): DateInterval
```

Slamdunk (changes requested):
Awesome :beers: 

Thank you very much for this, just a few edits and we can :ship: it

tests/Writer/XLSX/Helper/DateIntervalHelperTest.php
uncaught: Hmm, I don't think we can call a non-static method inside a static one, can we?

Slamdunk: That's the reason for https://github.com/openspout/openspout/pull/188#discussion_r1365225442

src/Writer/XLSX/Helper/DateIntervalHelper.php
uncaught: Sure we can, shall I bump the dependency in composer.json to `^0.27.6` ?

Slamdunk: Yes please

tests/Writer/XLSX/Helper/DateIntervalHelperTest.php
uncaught: But `provideDateIntervalToExcelCases()` is static, which is calling `invert()` !?

Slamdunk: Oh sorry, didn't see the origin.

Ok for the static then

src/Writer/XLSX/Helper/DateIntervalHelper.php
uncaught: Done

src/Reader/XLSX/Helper/DateIntervalFormatHelper.php
uncaught: Done

src/Reader/XLSX/Helper/DateIntervalFormatHelper.php
uncaught: Done

Slamdunk (approved):

PR #198, opened 2023-11-12
Fix to array returning formula instead of cell value

xaamin:
In [v4.16.0](https://github.com/openspout/openspout/releases/tag/v4.16.0) was introduced a breaking change due to returning formula values as-is so $row->toArray() in ExcelReader is not returning the cell value anymore. This PR adds a safety check to return the cell value on the toArray function of Row to mantain backwards compatibility with 4.x releases.

### DIFF
--- a/src/Common/Entity/Row.php
+++ b/src/Common/Entity/Row.php
@@ -7,6 +7,7 @@ namespace OpenSpout\Common\Entity;
 use DateInterval;
 use DateTimeInterface;
 use OpenSpout\Common\Entity\Style\Style;
+use OpenSpout\Common\Entity\Cell\FormulaCell;
 
 final class Row
 {
@@ -135,7 +136,9 @@ final class Row
     public function toArray(): array
     {
         return array_map(static function (Cell $cell): null|bool|DateInterval|DateTimeInterface|float|int|string {
-            return $cell->getValue();
+            return $cell instanceof FormulaCell
+                ? $cell->getComputedValue()
+                : $cell->getValue();
         }, $this->cells);
     }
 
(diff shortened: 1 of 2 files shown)


Slamdunk:
#169 has been a bugfix, not a BC break.

If you need the computed value, you should iterate over `getCells()` manually.

xaamin:
I considered that as a breaking change, I have a project deployed to production and after update the semver compatible it stopped working so after digging into the source code I noticed some tags after [v4.14.0](https://github.com/openspout/openspout/releases/tag/v4.14.0) someone changed the behaviour of FormulaCell so instead of returning the cell value it became retuning the cell formula.

xaamin:
What's your recommendation for this? I didn´t found any tests related to this so the broken change was silently skipped. Should I include bug fix in the PR description? I´m trying to understand why the PR was closed.

Slamdunk:
If you want to repropose this, I would merge the test part with the current behaviour, but without the changes over `src/Common/Entity/Row.php`, as I'm not going to release any revert about it.

xaamin:
If you merge just the test part with the current behaviour my test will not pass, the current vehaviour of [v4.19.0](https://github.com/openspout/openspout/releases/tag/v4.19.0) is different from [v4.14.0](https://github.com/openspout/openspout/releases/tag/v4.14.0) while they are considered semver compatible, I mean no breaking changes are expected.

The normal behaviour of my current deployed code with v4.14.0 vs v4.19.0 is explained with the next example:

** Having a three cells in a worksheet being A1: 5.50, B1: 10, C1: "=A1+B1" **

Behavior in v4.14.0 is while calling toArray on Row [5.50, 10, 15.50]
Behavior in v4.19.0 is while calling toArray on Row [5.50, 10, "=A1+B1"]

So they're not compatible at all, that's what the PR is intended for, but if openspout will considered this a valid semver change then I agree, my changes doesn´t make sense if this is considered the new behaviour.

PR #200, opened 2023-11-17
CSV: add support for DateTime and DateInterval objects

### LINKED ISSUE #199: Row::fromValues return array of cells which cannot be written by CSVWriter
Maybe I did not get correctly API and how to use it. But there is strange behavior when I have array of different objects and use Row::fromValues.

I have array like this one:
```
array(8) {
  ["firstName"]=>
  string(10) "John"
  ["lastName"]=>
  string(7) "Doe"
  ["created"]=>
  object(DateTime)#2068 (3) {
    ["date"]=>
    string(26) "2023-11-16 09:21:25.000000"
    ["timezone_type"]=>
    int(3)
    ["timezone"]=>
    string(3) "UTC"
  }
  ["notes"]=>
  NULL
 }
```

If I use **OpenSpout\Common\Entity\Row::fromValues** I got array of 4 cells and one of them is **DateTimeCell**. When I tried to write these cells by OpenSpout\Writer\XLSX\Writer everything is OK. But when I try to write these cells by OpenSpout\Writer\CSV\Writer it will throw exception because addRowToWriter [does not support DateTimeCell](https://github.com/openspout/openspout/blob/c0ece0d8c5d493fba6913c5c73d70887da249127/src/Writer/CSV/Writer.php#L52C9-L58C30).

Is it intended? If so why is only one method for converting data to cells instead of several different method or one which will return cells according to format (CSV, XLSX).

blazekv:
fixes #199


src/Writer/CSV/Options.php
```diff
@@ -9,6 +9,8 @@ final class Options
     public string $FIELD_DELIMITER = ',';
     public string $FIELD_ENCLOSURE = '"';
     public bool $SHOULD_ADD_BOM = true;
+    public string $DATE_TIME_FORMAT = DATE_ATOM;
+    public string $DATE_INTERVAL_FORMAT = '%y years, %m months, %d days, %h hours, %i minutes, %s seconds, %f microseconds';
```
Slamdunk: Thank you very much for investing your time in this library.

Regarding the formatting, I would like to avoid this time of customization because it doesn't preserve the possibility to easily read the CSV back with the reader.

So my proposal is to:

1. **not** allow any custom format for these fields; if the user want a custom format, simply iterate over the row cells and manually convert `DateTimeCell|DateIntervallCell` into a `StringCell`
2. default and only format for `DateTimeCell` is `DATE_ATOM`
3. default and only format for `DateIntervalCell` is `ISO-8601`, which doesn't come natively with PHP but [it can be easily coded](https://stackoverflow.com/questions/33787039/format-dateinterval-as-iso8601), see https://3v4l.org/VFN7N

Slamdunk (changes requested):

src/Writer/CSV/Options.php
blazekv: And how does it work excactly? I thought that if you want to read it back it will be string anyway because there is no way how to recognized that some text in CSV is date or date interval except of testing if text can be parsed via these formats and this is not happening right now isnt it?

I can change it if you are sure it is correct way, but from my perspective it make sense to have option to customize writer like this.

Slamdunk: > I thought that if you want to read it back it will be string anyway because there is no way how to recognized that some text in CSV is date or date interval except of testing if text can be parsed via these formats and this is not happening right now

You are indeed correct, but with `DATE_ATOM` and `ISO-8601` formats the user can easily translate the texts back to the corresponding PHP object

> I can change it if you are sure it is correct way, but from my perspective it make sense to have option to customize writer like this.

Please do it without customization options: we will add them later if really needed

blazekv: I remove options however, there is error from static analysis and I dont know why. It rejects that new types. It does not make sense to me. Because it prints that another type is expected.

```
::error file=src/Writer/CSV/Writer.php,line=,col=0::Ignored error pattern #^Parameter \#1 \$callback of function array_map expects \(callable\(OpenSpout\\Common\\Entity\\Cell\)\: mixed\)\|null, Closure\(OpenSpout\\Common\\Entity\\Cell\\BooleanCell\|OpenSpout\\Common\\Entity\\Cell\\EmptyCell\|OpenSpout\\Common\\Entity\\Cell\\FormulaCell\|OpenSpout\\Common\\Entity\\Cell\\NumericCell\|OpenSpout\\Common\\Entity\\Cell\\StringCell\)\: string given\.$# in path /var/www/html/src/Writer/CSV/Writer.php was not matched in reported errors.
::error file=src/Writer/CSV/Writer.php,line=52,col=0::Parameter #1 $callback of function array_map expects (callable(OpenSpout\Common\Entity\Cell): mixed)|null, Closure(OpenSpout\Common\Entity\Cell\BooleanCell|OpenSpout\Common\Entity\Cell\DateIntervalCell|OpenSpout\Common\Entity\Cell\DateTimeCell|OpenSpout\Common\Entity\Cell\EmptyCell|OpenSpout\Common\Entity\Cell\FormulaCell|OpenSpout\Common\Entity\Cell\NumericCell|OpenSpout\Common\Entity\Cell\StringCell): string given.
```
Changing that parameter only to Cell will bring another issues with return type.

Do you have any idea how to deal with it?

docs/documentation.md
```diff
@@ -37,6 +37,17 @@ $options->SHOULD_ADD_BOM = false;
 $writer = new Writer($options);
 ```
 
+You can specify how DateTime and DateInterval objects should be formatted. Writer calls the standard [DateTime format function](https://www.php.net/manual/en/datetime.format.php) or [DateInterval format function](https://www.php.net/manual/en/dateinterval.format.php) and uses the format parameter from the options.
```
Slamdunk: To be removed

tests/Writer/CSV/WriterTest.php
```diff
@@ -106,6 +108,32 @@ public function testWriteShouldSupportBooleanValues(): void
...
+    {
+        $now = new DateTime();
+        $options = new Options();
+        $allRows = $this->createRowsFromValues([
+            [$now],
+        ]);
+        $writtenContent = $this->writeToCsvFileAndReturnWrittenContent($allRows, 'csv_with_datetime_values.csv');
+        $writtenContent = $this->trimWrittenContent($writtenContent);
+
+        self::assertSame($now->format(DATE_ATOM), $writtenContent);
```
Slamdunk: In order for tests to be reliable, assertions should be done on hardcoded strings, so eventual bugs don't propagate.
Fix the date and time of the `DateTime` object and assert the output accordingly, doing something like:

```php
$now = new DateTime('2023-11-20 10:34:21');
// [...]
self::assertSame('2023-11-20T10:34:21+00:00', $writtenContent);
```

tests/Writer/CSV/WriterTest.php
```diff
@@ -106,6 +108,32 @@ public function testWriteShouldSupportBooleanValues(): void
...
+    {
+        $dateInterval = new DateInterval('P2Y3M10DT2H10M15S');
+        $options = new Options();
+        $allRows = $this->createRowsFromValues([
+            [$dateInterval],
+        ]);
+        $writtenContent = $this->writeToCsvFileAndReturnWrittenContent($allRows, 'csv_with_dateinterval_values.csv');
+        $writtenContent = $this->trimWrittenContent($writtenContent);
+
+        self::assertSame(implode('', [$options->FIELD_ENCLOSURE, $dateInterval->format('P%yY%mM%dDT%hH%iM%sS%fF'), $options->FIELD_ENCLOSURE]), $writtenContent);
```
Slamdunk: Same here:

```suggestion
        self::assertSame(implode('', [$options->FIELD_ENCLOSURE, 'P2Y3M10DT2H10M15S', $options->FIELD_ENCLOSURE]), $writtenContent);
```

Slamdunk (changes requested):

docs/documentation.md
blazekv: Removed

tests/Writer/CSV/WriterTest.php
blazekv: Fixed

tests/Writer/CSV/WriterTest.php
blazekv: Fixed

Slamdunk (approved):

Slamdunk:
Thank you very much :muscle:

Issue #129, opened 2023-01-03
XLSX date format incorrect when including escaped characters


harry-ybim:
I have an XLSX file with dates formatted: 
yyyy\-mm\-dd\Thh:mm:ss

This is incorrectly converted to a PHP date format: 
Y-m-dtH:i:s

The lowercase t should be maintained as an escaped uppercase T, like this: 
\T

The code producing this comes from here:
https://github.com/openspout/openspout/blob/e236da4d3ba365b6fb35803233a6f938497ca496/src/Reader/XLSX/Helper/DateFormatHelper.php#L82
https://github.com/openspout/openspout/blob/e236da4d3ba365b6fb35803233a6f938497ca496/src/Reader/XLSX/Helper/DateFormatHelper.php#L85

harry-ybim:
I've added a pull request here: https://github.com/openspout/openspout/pull/130

PR #204, opened 2023-11-22
Fix reading date formats that include T for ISO 8601

harry-ybim:
A new pull request for this issue: https://github.com/openspout/openspout/issues/129

### DIFF
--- a/src/Reader/XLSX/Helper/DateFormatHelper.php
+++ b/src/Reader/XLSX/Helper/DateFormatHelper.php
@@ -78,11 +78,19 @@ final class DateFormatHelper
                 continue;
             }
 
-            // Make sure all characters are lowercase, as the mapping table is using lowercase characters
-            $transformedPart = strtolower($dateFormatPart);
+            // Do not set \T to lower case or remove the slash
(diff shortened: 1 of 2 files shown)


src/Reader/XLSX/Helper/DateFormatHelper.php
```diff
@@ -78,11 +78,19 @@ public static function toPHPDateFormat(string $excelDateFormat): string
...
+            $dateFormatPartSections = explode('\T', $dateFormatPart);
+            foreach($dateFormatPartSections as &$dateFormatPartSection) {
 
-            // Remove escapes related to non-format characters
-            $transformedPart = str_replace('\\', '', $transformedPart);
+                // Make sure all characters are lowercase, as the mapping table is using lowercase characters
+                $dateFormatPartSection = strtolower($dateFormatPartSection);
+
+                // Remove escapes related to non-format characters
+                $dateFormatPartSection = str_replace('\\', '', $dateFormatPartSection);
```
Slamdunk, line 89: Infection says this is not tested, may I ask you why you wrote this line of code?

Slamdunk (changes requested):

PR #205, opened 2023-11-23
Allow passing options to Reader-/WriterFactory

BreyndotEchse:
Example:
```php
OpenSpout\Reader\Common\Creator\ReaderFactory::createFromFile($path, [
    'SHOULD_FORMAT_DATES' => false,
    'SHOULD_PRESERVE_EMPTY_ROWS' => false,
    'FIELD_DELIMITER' => ';',
    'FIELD_ENCLOSURE' => '\'',
]);
```

### DIFF
--- a/src/Reader/CSV/Options.php
+++ b/src/Reader/CSV/Options.php
@@ -12,4 +12,20 @@ final class Options
     public string $FIELD_DELIMITER = ',';
     public string $FIELD_ENCLOSURE = '"';
     public string $ENCODING = EncodingHelper::ENCODING_UTF8;
+
+    /**
+     * @param array<string, mixed> $options Array of options
+     *
+     * @return static
+     */
+    public static function fromArray(array $options): static
+    {
+        $self = new self();
+        $self->SHOULD_PRESERVE_EMPTY_ROWS = $options['SHOULD_PRESERVE_EMPTY_ROWS'] ?? $self->SHOULD_PRESERVE_EMPTY_ROWS;
+        $self->FIELD_DELIMITER = $options['FIELD_DELIMITER'] ?? $self->FIELD_DELIMITER;
+        $self->FIELD_ENCLOSURE = $options['FIELD_ENCLOSURE'] ?? $self->FIELD_ENCLOSURE;
+        $self->ENCODING = $options['ENCODING'] ?? $self->ENCODING;
+
+        return $self;
+    }
 }
--- a/src/Reader/Common/Creator/ReaderFactory.php
+++ b/src/Reader/Common/Creator/ReaderFactory.php
@@ -6,9 +6,12 @@ namespace OpenSpout\Reader\Common\Creator;
 
 use OpenSpout\Common\Exception\IOException;
 use OpenSpout\Common\Exception\UnsupportedTypeException;
+use OpenSpout\Reader\CSV\Options as CSVReaderOptions;
 use OpenSpout\Reader\CSV\Reader as CSVReader;
+use OpenSpout\Reader\ODS\Options as ODSReaderOptions;
 use OpenSpout\Reader\ODS\Reader as ODSReader;
 use OpenSpout\Reader\ReaderInterface;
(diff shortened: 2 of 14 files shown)


Slamdunk:
Why not setting the properties *after* the initialization, retaining all the goodness of strict types?

```php
$object = OpenSpout\Reader\Common\Creator\ReaderFactory::createFromFile($path);
$object->SHOULD_FORMAT_DATES = false;
```

BreyndotEchse:
1. You cannot change options of any reader for now. `Reader::$options` is `private readonly` and there is no getter (unlike writers: `getOptions()`).
2. You do not know, which reader will be returned by the factory. If it is a CSV-reader instance you do not want to set a value for the undeclared $SHOULD_FORMAT_DATES property. As of PHP 8.2 dynamic properties are deprecated. So you would have to do something like this but that would render the whole factory useless:
```php
$reader = OpenSpout\Reader\Common\Creator\ReaderFactory::createFromFile($path);
if ($reader instanceof X) {
    $reader->getOptions()->foo = 'bar';
} elseif ($reader instanceof Y) {
    //...
} elseif ($reader instanceof Z) {
    //...
}
```
3. All readers and writers accept options in their constructors so it is a matter of stringency to allow options to get passed to the factory.
4. In my opinion factories should return always ready-to-use objects - some rare or very specific use cases left out.

Slamdunk:
> 1. You cannot change options of any reader for now. `Reader::$options` is `private readonly` and there is no getter (unlike writers: `getOptions()`).

Right, sorry, I haven't looked well enough the original source code.

This behaviour is indeed by design: the less mutable code there is in the architecture, the easier it is to maintain.

> 2. You do not know, which reader will be returned by the factory. If it is a CSV-reader instance you do not want to set a value for the undeclared $SHOULD_FORMAT_DATES property.

If you do not know which reader will be returned, how do you know which parameters to pass to the factory?
If you still need the `if {} elseif {} else {}` code, than you don't need the factory at all, and the factory would be rightfully useless.

> 3. All readers and writers accept options in their constructors so it is a matter of stringency to allow options to get passed to the factory.

But each has its unique dependencies, you cannot generlize the signature.

> 4. In my opinion factories should return always ready-to-use objects - some rare or very specific use cases left out.

You are right, but that's true for good factories that return a known type.
Those are not factories but rather some simple guessers. I don't like them, I just inherited them with the code base.

BreyndotEchse:
> If you do not know which reader will be returned, how do you know which parameters to pass to the factory? If you still need the `if {} elseif {} else {}` code, than you don't need the factory at all, and the factory would be rightfully useless.

With this commit you can pass any option of any reader/writer to the factory. The respective `***Options::fromArray` factory method will only use eligible options following the SoC principle.

> But each has its unique dependencies, you cannot generlize the signature.

The whole purpose of factories is to provide polymorphism by generalizing the creation of its' "products". That does not exclude the creation and injection of dependent objects.

> You are right, but that's true for good factories that return a known type. Those are not factories but rather some simple guessers. I don't like them, I just inherited them with the code base.

I would say those are [Simple Factories](https://github.com/bethrobson/Head-First-Design-Patterns/blob/master/src/headfirst/designpatterns/factory/pizzas/SimplePizzaFactory.java) and I don't think there is anything wrong with it.

If you want to get rid of the factories I don't have a problem with it to get tagged as `@deprecated` and removed in the next major release. But as long as it is part of the code base and people rely on it, it should provide a properly functioning way to return a ready-to-use reader or writer.

Slamdunk:
> With this commit you can pass any option of any reader/writer to the factory.

I want robust code. If we allow `any` type to be valid, the user will never be able to know if something passed to the factory is useful or not at compile time, but only at runtime the check can be enforced, which is bad practice at best, a security threat at worst.

As far as I can tell, the only acceptable PR against `ReaderFactory` and `WriterFactory` is to make them `@deprecated`.

Slamdunk:
Closed by https://github.com/openspout/openspout/pull/212

PR #208, opened 2023-12-15
Add support for headerFooter, fitToHeight/Width, printTitleRows (xlsx only)

tekku-taro:
### headerFooter

Header and footer can be set via Options::setHeaderFooter()
```
$options = new Options();
$options->setHeaderFooter(new HeaderFooter(
    'oddHeader',
    'oddFooter',
    'evenHeader',
    'evenFooter',
    true  // differentOddEven, default value is false
));
```
### fitToHeight/Width

Add new arguments for fitToHeight and fitToWidth to the PageSetup constructor.
```
$options = new Options();
$options->setPageSetup(new PageSetup(
    PageOrientation::LANDSCAPE,
    PaperSize::A4,
    0, // ?int fitToHeight
    1  // ?int fitToWidth
));
```

### printTitleRows 

Print title rows can be set at the sheet level for XLSX writers.

```
$sheet = $writer->getCurrentSheet();
$sheet->setPrintTitleRows('$2:$2');
```

### Examples (preview images)

image1
image2
image3

### DIFF
--- a/src/Writer/XLSX/Helper/FileSystemHelper.php
+++ b/src/Writer/XLSX/Helper/FileSystemHelper.php
@@ -310,7 +310,7 @@ final class FileSystemHelper implements FileSystemWithRootFolderHelperInterface
     public function createContentFiles(Options $options, array $worksheets): self
     {
         $allMergeCells = $options->getMergeCells();
-
+        $pageSetup = $options->getPageSetup();
         foreach ($worksheets as $worksheet) {
             $contentXmlFilePath = $this->getXlWorksheetsFolder().\DIRECTORY_SEPARATOR.basename($worksheet->getFilePath());
             $worksheetFilePointer = fopen($contentXmlFilePath, 'w');
@@ -332,7 +332,9 @@ final class FileSystemHelper implements FileSystemWithRootFolderHelperInterface
                 fwrite($worksheetFilePointer, '<sheetPr filterMode="false"><pageSetUpPr fitToPage="false"/></sheetPr>');
                 fwrite($worksheetFilePointer, sprintf('<dimension ref="%s"/>', $range));
             }
-
+            if (isset($pageSetup) && $pageSetup->fitToPage) {
+                fwrite($worksheetFilePointer, '<sheetPr><pageSetUpPr fitToPage="true"/></sheetPr>');
+            }
             if (null !== ($sheetView = $sheet->getSheetView())) {
                 fwrite($worksheetFilePointer, '<sheetViews>'.$sheetView->getXml().'</sheetViews>');
             }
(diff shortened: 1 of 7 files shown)


src/Writer/XLSX/Helper/FileSystemHelper.php
```diff
@@ -329,7 +332,9 @@ public function createContentFiles(Options $options, array $worksheets): self
                 fwrite($worksheetFilePointer, '<sheetPr filterMode="false"><pageSetUpPr fitToPage="false"/></sheetPr>');
                 fwrite($worksheetFilePointer, sprintf('<dimension ref="%s"/>', $range));
             }
-
+            if (isset($pageSetup) && $pageSetup->fitToPage) {
+                fwrite($worksheetFilePointer, '<sheetPr><pageSetUpPr fitToPage="true"/></sheetPr>');
```
Slamdunk, line 339: `sheetPr` tag is already written within the autofilter `if` few lines above: we need to handle the case where both are used

src/Writer/XLSX/Options/PageSetup.php
```diff
@@ -6,8 +6,16 @@
 
 final class PageSetup
 {
+    public bool $fitToPage = false;
```
Slamdunk:
```suggestion
    public readonly bool $fitToPage = false;
```

Slamdunk (changes requested):

src/Writer/XLSX/Helper/FileSystemHelper.php
tekku-taro: From what I understand, when both autofilter and fitToPage are turned on, the fitToPage attribute will be true, right?

src/Writer/XLSX/Options/PageSetup.php
tekku-taro: sorry, this can't be accepted. The following error occurs:
```
$fitToPage cannot have default value
```
Besides, fitToPage value depends on fitToHeight and fitToWidth, which is passed to the constructor.

Slamdunk: Just remove the default value and set it in the constructor as

```php
$this->fitToPage = null !== $fitToHeight || null !== $fitToWidth;
```

src/Writer/XLSX/Helper/FileSystemHelper.php
tekku-taro: fixed it to handle the case.
[b826464](https://github.com/openspout/openspout/pull/208/commits/b82646439dc1837339e5c404867af5c0ee05d0c6)

src/Writer/XLSX/Options/PageSetup.php
tekku-taro: I fixed it.

Slamdunk (approved):

Slamdunk:
Thank you @tekku-taro :muscle:

PR #228, opened 2024-03-03
The cell style is optional

lav45:


src/Common/Entity/Cell.php
```diff
@@ -22,44 +22,41 @@ abstract class Cell
 
     private Style $style;
 
-    public function __construct(?Style $style)
+    public function withStyle(Style $style): static
```
Slamdunk: I like this change, but this is a https://github.com/openspout/openspout/labels/BC%20Break: please restore the constructor and add a `@deprecated` doc for the parameter, so we can merge this in the current major

Slamdunk (changes requested):

ziegenberg:
@Slamdunk, what would be the best course of action in light of these changes?

Slamdunk:
If the PR gets rebased and conflicts resolved, I'm ready to merge this :+1:

ziegenberg:
> If the PR gets rebased and conflicts resolved, I'm ready to merge this :+1: 

So you don't want to staggered process with deprecation message first?

PR #230, opened 2024-03-21
Enhance Row::fromValues can accept Cell object (StringCell)

muhtaromzain:
I was wondering if we could accept only for specific cell with style using Cell::fromValue using Row::fromValues as following example data:

$values = [
       "string",
       "string",
       Cell::fromValue("string", Style)
];

Row::fromValues($values);

I was concerning about if we have a huge number of data, it would be unnecessary to put every single data with function Cell::fromValue if the data using default style.

### DIFF
--- a/src/Common/Entity/Row.php
+++ b/src/Common/Entity/Row.php
@@ -6,6 +6,7 @@ namespace OpenSpout\Common\Entity;
 
 use DateInterval;
 use DateTimeInterface;
+use OpenSpout\Common\Entity\Cell\StringCell;
 use OpenSpout\Common\Entity\Style\Style;
 
 final class Row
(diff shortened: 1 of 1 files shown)


Slamdunk:
I don't get what's the underlying problem here :thinking:

muhtaromzain:
> I don't get what's the underlying problem here 🤔

Let me give you an example, Here I have an array with 50 lengths:

$values = [
     'string',
     'string',
     'string',
     'string',
     'string',
     'string',
     'string',
     'string',
     ...
]; 

Lets give style only on 1 specific cell like this.

$values = [
     'string',
     'string',
     'string',
     'string',
     'string',
     'string',
     Cell::fromValue('string', Style),
     'string',
     ...
]; 

Row::fromValues($values);

This would cause error since fromValues would not accept StringCell, and only can be achieved by using Cell:fromValue on each data, like this.

$values = [
      Cell::fromValue('string'),
      Cell::fromValue('string'),
      Cell::fromValue('string'),
      Cell::fromValue('string'),
      Cell::fromValue('string'),
      Cell::fromValue('string'),
      Cell::fromValue('string', Style),
      Cell::fromValue('string'),
     ...
]; 

$rowValues = new Row($values);

Instead of doing this, it would be nice if we can customize like the first one, if we already have a bunch of function write using Row:fromValues build like this below.

$values = [
     'string',
     'string',
     'string',
     'string',
     'string',
     'string',
     'string',
     'string',
     ...
]; 

Row::fromValues($values);

It would be nice to have only update only specific row like the first one. Hope my explanation helps. Thank you

Slamdunk:
Unless a specific performance issue appears, type consistency is much more important than usage convenience.

PR #235, opened 2024-04-25
Create row from values with per cell styling

jschram:
Allows to specify an array with styles per column when calling `Row::fromValues`.
This could for instance be used to mark certain cells as date:

```php
  Row::fromValues(
    cellValues: $values, 
    columnStyles: [
      0 => (new Style[])->setFormat('dd-mm-yyyy'),
    ]
  )
```

### DIFF
--- a/src/Common/Entity/Row.php
+++ b/src/Common/Entity/Row.php
@@ -38,9 +38,21 @@ final class Row
 
     /**
      * @param list<null|bool|DateInterval|DateTimeInterface|float|int|string> $cellValues
-     * @param Style[]                                                         $columnStyles
      */
-    public static function fromValues(array $cellValues = [], ?Style $rowStyle = null, array $columnStyles = []): self
+    public static function fromValues(array $cellValues = [], ?Style $rowStyle = null): self
+    {
+        $cells = array_map(static function (null|bool|DateInterval|DateTimeInterface|float|int|string $cellValue): Cell {
+            return Cell::fromValue($cellValue);
+        }, $cellValues);
+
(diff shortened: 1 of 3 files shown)


jschram:
Hi @Slamdunk! Is there anything I can do to aid you in reviewing this PR? Please let me know if you need further explanation!

src/Common/Entity/Row.php
```diff
@@ -38,12 +38,13 @@ public function __construct(array $cells, ?Style $style = null)
 
     /**
      * @param list<null|bool|DateInterval|DateTimeInterface|float|int|string> $cellValues
+     * @param Style[]                                                         $columnStyles
```
Slamdunk: Currently this cannot be merged *as-is*: the `$cellValues` argument is typed as `list<>`, and relying on `$columnStyles` to have the same keys can't be requested nor enforced.

But a new API is welcome, so please:

1. Add a new `fromValuesWithStyle` API with both `$cellValues` and `$columnStyles` typehinted as `array<array-key, [...]>`
2. Add tests for the new API
3. Add brief mention to the docs for the new API

Slamdunk (changes requested):

src/Common/Entity/Row.php
jschram: Thanks for the suggestion! I've made the required changes in this commit: [ac52867](https://github.com/openspout/openspout/pull/235/commits/ac5286704294654d7e2ae042ea587cb2b129e5f5)

Slamdunk (approved):

PR #239, opened 2024-05-20
Raise exception when file is missing and `realpath` returns `false`

### LINKED ISSUE #238: Unable to open/rewind RowIterator when XLXS is a temp-file
We download file as stream, save that into tmp file

```php
        /** @var resource $file */
        $file = tmpfile();
        fwrite($file, $data);
        $meta = stream_get_meta_data($file);

        $reader = new Reader();
        $reader->open($meta['uri']);

        $sheetIterator = $reader->getSheetIterator();
        $sheetIterator->rewind();

        return $sheetIterator->current();
```

```php
        $rowIterator = $spreadsheet->getRowIterator();
        $rowIterator->rewind();
```

`fileExistsWithinZip` has "wrong" expression

image

`/zip:\/\/([^#]+)#(.*)/` does not match `zip://#xl/worksheets/sheet1.xml`

----

After debugging a bit more looks like path is not "concated" in `getRealPathURIForFileInZip`

image

it's empty because `realpath` may return FALSE and when converted to string it's a `empty string` which means that tmp-file is closed and deleted.

image

oleg-andreyev:
This could occur when reading from tmp-file which is closed somewhere in the middle

fixes #238

### DIFF
--- a/src/Reader/Wrapper/XMLReader.php
+++ b/src/Reader/Wrapper/XMLReader.php
@@ -58,7 +58,7 @@ final class XMLReader extends \XMLReader
             throw new IOException("Could not open {$zipFilePath} for reading! File does not exist.");
         }
 
-        return self::ZIP_WRAPPER. $realpath .'#'.$fileInsideZipPathWithoutLeadingSlash;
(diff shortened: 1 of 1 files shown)


Slamdunk:
Thank you :muscle:

PR #244, opened 2024-05-31
[XLSX] Add textRotation to cell style

TekMate8:


Slamdunk:
Hi, can you add a test please?

TekMate8:
Hi! I'm new at adding pull requests to github repo. I have added some tests to the StyleMergerTest class, I hope it's okay. If not, let me know what else should I do or point me to an example. Thanks.

src/Common/Entity/Style/Style.php
```diff
@@ -383,6 +389,28 @@ public function hasSetWrapText(): bool
...
 
+    public function textRotation(): int
+    {
+        return $this->textRotation;
+    }
+
+    /**
+     * @param int $rotation Rotate text
+     */
+    public function setTextRotation(int $rotation = 0): self
```
Slamdunk:
```suggestion
    public function setTextRotation(int $rotation): self
```

src/Common/Entity/Style/Style.php
```diff
@@ -383,6 +389,28 @@ public function hasSetWrapText(): bool
...
+    {
+        $this->textRotation = $rotation;
+        $this->hasSetTextRotation = true;
+        $this->isEmpty = false;
```
Slamdunk, line 404: Mutation testing tells us this line is not tested, can we remove it?

src/Common/Entity/Style/Style.php
```diff
@@ -383,6 +389,28 @@ public function hasSetWrapText(): bool
...
+
+        return $this;
+    }
+
+    public function hasSetTextRotation(): bool
+    {
+        return $this->hasSetTextRotation;
```
Slamdunk, line 411: `0` seems to be the default value: can we remove the `hasSetTextRotation` property and just check if `0 !== $this->textRotation`?

Slamdunk (changes requested):

src/Common/Entity/Style/Style.php
TekMate8: Yes.

Slamdunk (approved):

Slamdunk:
Thank you

Issue #252, opened 2024-06-20
Allow (optional) file locking


inmanturbo:
I don't see flock() called anywhere in this code base. Would the maintainers be open to allowing an option for locking the file for writing? To help avoid race conditions when another process is writing to the same file.

e.g. something pseudo-like

```php
$resource = fopen($this->outputFilePath, 'w');

if  ($this->options->LOCK_FILE_FOR_WRITING && !flock($resource, LOCK_EX)) {
    // handle error;
}
```

Slamdunk:
Sure 👍

PR #253, opened 2024-06-20
allow optional flock of file when writing

inmanturbo:
https://github.com/openspout/openspout/issues/252

The simplest way it seems is to call `flock($resource, LOCK_EX)` in `AbstractWriter::openToFile`,  then `flock($this->filePointer, LOCK_UN)` in `AbstractWriter::close` . However the abstract class does not have an options property.

I've added `public bool $shouldLockFileForWriting = false;` to the `AbstractWriter`, then set this property based on `Options` in the `Writer`s' constructors. Of course the default is false so it won't change the behavior without first setting the option to `true`.

I'm not sure if this is how the maintainers would want this implemented, so I'm opening as a draft to start.  Open to suggestions and review.

Example usage:

```php

$options = new OpenSpout\Writer\CSV\Options();

$options->SHOULD_LOCK_FILE_FOR_WRITING = true;

$writer = new OpenSpout\Writer\CSV\Writer($options);
```

### DIFF
--- a/src/Writer/AbstractWriter.php
+++ b/src/Writer/AbstractWriter.php
@@ -28,6 +28,9 @@ abstract class AbstractWriter implements WriterInterface
     /** @var 0|positive-int */
     private int $writtenRowCount = 0;
 
+    /** @var false|bool */
+    public bool $shouldLockFileForWriting = false;
+
     final public function openToFile($outputFilePath): void
     {
         $this->outputFilePath = $outputFilePath;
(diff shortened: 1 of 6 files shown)


imflyou:
Does the option `SHOULD_LOCK_FILE_FOR_WRITING` default value should be `true` ?

inmanturbo:
> Does the option `SHOULD_LOCK_FILE_FOR_WRITING` default value should be `true` ?

I don't know if it should. The package has never locked files before, so users of the package likely would not expected it to lock (or block) by default.

inmanturbo:
I still need to find a good way write a test to test for `flock()` when I get a chance.

probably with `flock($handle, LOCK_EX|LOCK_NB, $wouldblock)` ? 

Also will need to update Docs. Still a draft ...

inmanturbo:
For my own case I'm using this in a `Laravel` app, and as a workaround for now I'm using [atomic locks](https://laravel.com/docs/11.x/cache#atomic-locks)

PR #257, opened 2024-07-12
Fix: always write sheet `dimension` tag even when there's no autofilter

aprat84:
The `dimension` tag for each sheet tells the cell range that sheet data is using, but it has no relation with autofilter.

The autofilter tag has its own cell range.

Other libraries rely on the `dimension` tag to get the cell range fast, otherwise they have to parse whole file, and can be slow on huge files.

Added a simple test.


Slamdunk (approved):

Slamdunk:
Thank you

aprat84:
Thanks for the fast merge!

As a note: the starting range is hardcoded at `A1`, cause this library does not have a functionality of writing cells at a given coordinate. The only way to add data AFAIK is by calling `Writer::addRow()` or `Writer::addRows()`, which always starts at `A1`.

PR #260, opened 2024-07-23
Add XLSX option whether formulas should be read as FormulaCells

karpilin:


tests/Reader/XLSX/ReaderTest.php
```diff
@@ -562,12 +562,13 @@ public function testReadShouldBeAbleToProcessEmptySheets(): void
...
     }
 
-    public function testReadShouldSkipFormulas(): void
+    public function testReadShouldReadFormulas(): void
     {
         $resourcePath = TestUsingResource::getResourcePath('sheet_with_formulas.xlsx');
 
         $allRows = [];
         $options = new Options();
+        $options->SHOULD_READ_FORMULAS = true;
```
Slamdunk, line 571: The primary goal of this library is to be idempotent: reading and re-writing a file should yield the same result.

The default value here should be false, hence this diff must be reverted.

Slamdunk (changes requested):

tests/Reader/XLSX/ReaderTest.php
karpilin: Thanks for looking into this. I agree with you and no longer think this extra option need to be added at all. I'll close the PR.

PR #275, opened 2024-10-04
Implement collapsible columns for XLSX

kamilrzany:
**Overview**
This PR introduces the ability to collapse selected columns in XLSX documents.

CleanShot 2024-10-04 at 16 05 05

**Usage**
```
$columns = [2, 3];
$sheet->setColumnAttributes(1, false, true, ...$columns);
```
It will group the 2nd and 3rd columns under the 1st outline level.

**Challenges**
I'm still unsure about the best way to implement this and would really appreciate your feedback and suggestions. If you think a different approach might be better, please let me know. I'm open to refactoring it if we can agree on a solid solution.

Initially, I implemented this by adding extra arguments to the `setColumnWidth()` and `setColumnWidthForRange()` methods, but then I realized that this might not be the best idea, as every user would need to update their codebase to adapt to the new method structure.

I considered also on creating extra method(s) like `setOutlineLevel()` and `setCollapsed()` directly in `AbstractOptions` class. However, implementing this may be challenging, because we would need to add options to the existing `ColumnWidth` objects in `$this->COLUMN_WIDTHS`.

**Next Steps**
- [ ] Update the documentation
- [ ] Implement it for `AbstractOptions.php`
- [ ] Add more tests
- [ ] Refactor code to follow all code standards

### DIFF
--- a/src/Writer/Common/ColumnAttributes.php
+++ b/src/Writer/Common/ColumnAttributes.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\Common;
+
+/**
+ * @internal
+ */
+final class ColumnAttributes
+{
+    public function __construct(
+        public readonly ?int $outlineLevel = null,
+        public readonly bool $collapsed = false,
+        public readonly bool $hidden = false
+    ) {
+    }
+
+    public function getOutlineLevel(): ?int
+    {
+        return $this->outlineLevel;
+    }
+
+    public function setOutlineLevel(?int $outlineLevel): void
+    {
+        $this->outlineLevel = $outlineLevel;
+    }
+
+    public function isCollapsed(): bool
+    {
+        return $this->collapsed;
+    }
+
+    public function setCollapsed(bool $collapsed): void
+    {
+        $this->collapsed = $collapsed;
+    }
+
+    public function isHidden(): bool
+    {
+        return $this->hidden;
+    }
+
+    public function setHidden(bool $hidden): void
+    {
+        $this->hidden = $hidden;
+    }
+
+    public function isEmpty(): bool
+    {
+        return null === $this->getOutlineLevel()
+            && false === $this->isCollapsed()
+            && false === $this->isHidden();
+    }
+}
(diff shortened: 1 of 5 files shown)


kamilrzany:
Hey guys, I created a draft pull request so I could ask you about the best possible implementation of this feature. I've put my thoughts on it in the PR description in the Challenges section.

jhogervorst:
@Slamdunk Hey! 👋 We started using OpenSpout in our project. Being able to collapse columns (in XLSX files) is an important feature for us, and I think it is useful for other OpenSpout users as well.

We're currently using our fork of OpenSpout with this feature included. However, we'd prefer to get it merged, so it will be part of the official release and we don't need to keep using our own fork.

This current PR is not completely ready, because we were not sure what would be the best approach. My colleague @kamilrzany explained the challenges and some ideas in this PR's description.

Would you be able to take a look and let us know if you're open to merging this feature, and which approach would be preferred? We're happy to work a bit more on the PR, to make it nice according to the chosen approach.

I've sent a contribution through GitHub Sponsors to make up for your time 🙂

Slamdunk:
Hello, thank you for your sponsorship and the time you took to improve this library.

I researched the topic a bit more and I found that the `<col>` attribute in XLSX has been architectured with the false idea that all the attributes are related when instead they aren't.

This means for us that we can safely implement the properties unrelated one from each other, and only merge them once we need to write the `<col>` tag.

So the approach I'd take is the following:

1. Write different classes: `ColumnWidth` (already done), `ColumnHidden`, `ColumndOutlineLevel` and `ColumnCollapsed`
2. Expand the `Sheet` API to allow setting them separately
3. Merge them in the `<col>` attribute with distinct column ranges. This means that if the user asks for A+B to be hidden and B+C with `OutlineLevel>1`, we need 3 `<col>` tags (A, B and C), but instead for A+B hidden and A+B+C `OutlineLevel>1` only two `<col>` are needed (A+B and C)
4. Completely skip the `Options` class: it has been a mistake by me to consider the `ColumnWidth` for the `Workbook` and it will be removed from the next major release. All the attributes belong only to the `Worksheet`

jhogervorst:
@Slamdunk Thanks so much for your investigation and clear reply! I'll review this with @kamilrzany and we'll schedule some time for refactoring this in our planning. We'll update this PR when it's ready.

nurikabe:
This and collapsible rows would be nice.

Slamdunk:
Closed by https://github.com/openspout/openspout/pull/396

PR #283, opened 2024-11-23
Add XLSX Protection capabilities

freshleafmedia:
This PR adds support for Workbook and Worksheet protection to XLSX.

- [WorkbookProtection OOXML Spec](https://c-rex.net/samples/ooxml/e1/Part4/OOXML_P4_DOCX_workbookProtection_topic_ID0ECHK4.html?hl=workbookprotection)
- [WorkbookProtection MS Spec](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.workbookprotection?view=openxml-3.0.1)
- [SheetProtection OOXML Spec](https://c-rex.net/samples/ooxml/e1/Part4/OOXML_P4_DOCX_sheetProtection_topic_ID0ENIH5.html)
- [SpeetProtection MS Spec](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.sheetprotection?view=openxml-3.0.1)

### DIFF
--- a/src/Writer/XLSX/Options/SheetProtection.php
+++ b/src/Writer/XLSX/Options/SheetProtection.php
@@ -1,0 +1,285 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\XLSX\Options;
+
+final class SheetProtection
+{
+    public ?string $passwordHash = null;
+    public bool $lockSheet = false;
+    public bool $lockColumnInsert = false;
+    public bool $lockColumnDelete = false;
+    public bool $lockColumnFormatting = false;
+    public bool $lockRowInsert = false;
+    public bool $lockRowDelete = false;
+    public bool $lockRowFormatting = false;
+    public bool $lockAutoFilter = false;
+    public bool $lockSort = false;
+    public bool $lockCellFormatting = false;
+    public bool $lockLockedCellSelection = false;
+    public bool $lockUnlockedCellsSelection = false;
+    public bool $lockObjects = false;
+    public bool $lockHyperlinkInsert = false;
+    public bool $lockPivotTables = false;
+    public bool $lockScenarios = false;
+
+    /**
+     * @return $this
+     */
+    public function setPassword(string $password): self
+    {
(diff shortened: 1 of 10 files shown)


src/Writer/XLSX/Options/SheetProtection.php
```diff
@@ -0,0 +1,285 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\XLSX\Options;
+
+final class SheetProtection
```
Slamdunk:
```suggestion
final readonly class SheetProtection
```
1. Please mark the class as `readonly`
2. Make all the properties `public` so we can get rid of the setters/getters

src/Writer/XLSX/Options/WorkbookProtection.php
```diff
@@ -0,0 +1,105 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\XLSX\Options;
+
+final class WorkbookProtection
```
Slamdunk:
```suggestion
final readonly class WorkbookProtection
```
1. Please mark the class as `readonly`
2. Make all the properties `public` so we can get rid of the setters/getters

src/Writer/XLSX/Options/WorkbookProtection.php
```diff
@@ -0,0 +1,105 @@
...
+
+        return $this;
+    }
+
+    public function getXml(): string
+    {
+        return '<workbookProtection'.$this->getSheetViewAttributes().'/>';
+    }
+
+    private function createPasswordHash(string $password): string
```
Slamdunk: This seems copy-pasted from `SheetProtection`.
Please:

1. Move the hashing function into a separate class
2. Add a specific test with hardcoded password and hash result

Slamdunk (changes requested):

freshleafmedia:
@Slamdunk I have made the requested changes. :+1:

Slamdunk (approved):

PR #287, opened 2024-11-29
Restore `setCreator` API dropped by accident in a minor release

### LINKED ISSUE #286: v4.28.0 broken public api
I don't know if its intentional, but https://github.com/openspout/openspout/releases/tag/v4.28.0 removed public setCreator method from the XLSX Writer as part of https://github.com/openspout/openspout/pull/285/files and it seems very odd that this is done on a minor release ?
Or I've missed something in the recent updates

i would keep the setCreator method and simply make it update the underlying properties object (it's readonly, so it's a bit tricky if you keep it that way)

lekoala:
this will also require Properties to not be readonly

fixes https://github.com/openspout/openspout/issues/286

### DIFF
--- a/src/Writer/AbstractWriter.php
+++ b/src/Writer/AbstractWriter.php
@@ -137,8 +137,6 @@ abstract class AbstractWriter implements WriterInterface
         $this->isWriterOpened = false;
     }
 
-    abstract public function setCreator(string $creator): void;
-
     /**
      * Opens the streamer and makes it ready to accept data.
      *
--- a/src/Writer/AbstractWriter.php
+++ b/src/Writer/AbstractWriter.php
@@ -137,6 +137,8 @@ abstract class AbstractWriter implements WriterInterface
         $this->isWriterOpened = false;
     }
 
+    abstract public function setCreator(string $creator): void;
+
     /**
      * Opens the streamer and makes it ready to accept data.
      *
--- a/src/Writer/CSV/Writer.php
+++ b/src/Writer/CSV/Writer.php
@@ -28,6 +28,11 @@ final class Writer extends AbstractWriter
     {
         return $this->options;
     }
+    
+    public function setCreator(string $creator): void
+    {
+        // noop
+    }
 
     /**
      * Opens the CSV streamer and makes it ready to accept data.
--- a/src/Writer/CSV/Writer.php
+++ b/src/Writer/CSV/Writer.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace OpenSpout\Writer\CSV;
 
+use Exception;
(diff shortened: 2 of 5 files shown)


src/Writer/XLSX/Properties.php
```diff
@@ -4,7 +4,7 @@
 
 namespace OpenSpout\Writer\XLSX;
 
-final readonly class Properties
+final class Properties
```
Slamdunk: This is not needed, just overwrite the object with a new one

src/Writer/XLSX/Writer.php
```diff
@@ -35,6 +35,11 @@ public function getOptions(): Options
         return $this->options;
     }
 
+    public function setCreator(string $creator): void
```
Slamdunk, line 38: I guess we also need to restore this in the `WriterInterface`

Slamdunk (changes requested):

lekoala:
@Slamdunk kept it readonly (it's just a bit verbose imho), added back method to interface. instead of implementing in abstract, i kept an abstract method and added a noop function to the csv writer

src/Writer/AbstractWriter.php
```diff
@@ -137,6 +137,8 @@ final public function close(): void
         $this->isWriterOpened = false;
     }
 
+    abstract public function setCreator(string $creator): void;
```
Slamdunk: This is already enforced by the interface and can be dropped here

src/Writer/CSV/Writer.php
```diff
@@ -28,6 +28,11 @@ public function getOptions(): Options
     {
         return $this->options;
     }
+    
+    public function setCreator(string $creator): void
+    {
+        // noop
```
Slamdunk: May it would be better to raise an Exception message like `Method unsopported for CSV documents`

Slamdunk (changes requested):

lekoala:
@Slamdunk i've made the requested changes

i'm not planning on updating this pr further than that, it feels like you would spend as much time writing the comment as doing the change yourself which seem counterproductive. my proposal would be to merge as is and make any last changes as you see fit.

have a great day,

Slamdunk:
> it feels like you would spend as much time writing the comment as doing the change yourself which seem counterproductive

It depends on the definition of counterproductive.
I review PR on the open source projects I maintain to share my knowledge so that the knowledge itself can improve for everyone from anyone has better ideas and suggestions.

If you feel annoyed by my feedbacks, imagine how I feel having to deal with hundreds of people like you after I spent my free time in the pursuit of reciprocal help.

Slamdunk (approved):

PR #290, opened 2024-12-17
fix: handle veryHidden sheet state

blinkseb:
Hello!

While debugging an issue, I stumble across a XLSX document with a hidden sheet that was reported as visible by openspout. Looking at the raw XML, the sheet state is set to `veryHidden` instead of `hidden`. It seems that this state can only be set programmatically (https://learn.microsoft.com/en-us/office/troubleshoot/excel/hide-sheet-and-use-xlveryhidden).

Here's an extract raw XML of the document:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<workbook xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"/>
   [...]
    <sheets>
        <sheet name="Sheet1" sheetId="1" state="veryHidden" r:id="rId1" />
        <sheet name=" Liste Evenements" sheetId="4" r:id="rId2" />
    </sheets>
  [...]
</workbook>
```

I've added support for `veryHidden` state when computing the sheet visibility flag, and added a test case.

Thanks!


Slamdunk:
Thank you

PR #330, opened 2025-09-03
Fix coding standards

stof:
The latest version of php-cs-fixer enforces `null` to come last in union types in the configuration being used.


Slamdunk:
Thank you

PR #334, opened 2025-10-27
Make ODS  Vertical Alignment work ?

vayanes:
I propose these changes to make Vertical Alignment work for ODS.

In fact, the style for vertical alignment seems to have to  be in table-cell properties and not in paragraph-properties.

### DIFF
--- a/src/Reader/XLSX/Helper/CellHelper.php
+++ b/src/Reader/XLSX/Helper/CellHelper.php
@@ -48,7 +48,7 @@ final class CellHelper
         // Also, not using the pow() function because it's slooooow...
         switch ($columnLength) {
             case 1:
-                $columnIndex = self::columnLetterToIndexMapping[$columnLetters];
+                $columnIndex = self::columnLetterToIndexMapping[$columnLetters[0]];
 
                 break;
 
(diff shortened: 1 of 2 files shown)


Slamdunk (changes requested):
Hello, as you can see by the feedback given by the checks, the proposed code needs to:

1. Have coding standards fixed: just run `make` in you cli and you should be fine
2. Pass the tests: failing tests must be inspected, and the contributor have to decide what's the expected behavior in all the failing scenarios

Also, a new test should be added to ensure the new feature doesn't break in the future by other new changes

Slamdunk:
Hello @vayanes it seems to me you are making edits locally and pushing them here to see if the build goes green.
You don't need to wait: you can build everything locally on your computer by having `docker` installed and typing `make` in the repo root dir :wink:

vayanes:
Oui, j'ai essayé, mais j'avais des erreurs de docker.

De toute façon, j'arrête : je n'ai pas le temps de jouer aux devinettes.

Yes, I tried, but I had Docker errors.

Anyway, I'm giving up: I don't have time to play guessing games.

PR #340, opened 2025-12-09
Add Withers to Common/Entity classes

karpilin:
- chainable "with"/"without" methods
- move default values to public constants

### DIFF
--- a/src/Common/Entity/Cell/BooleanCell.php
+++ b/src/Common/Entity/Cell/BooleanCell.php
@@ -25,4 +25,34 @@ final readonly class BooleanCell extends Cell
     {
         return $this->value;
     }
+
+    public static function create(bool $value): self
+    {
+        return new self($value);
+    }
+
+    public function withValue(bool $value): self
+    {
+        return new self($value, $this->style, $this->comment);
+    }
+
+    public function withStyle(Style $style): self
+    {
+        return new self($this->value, $style, $this->comment);
+    }
+
+    public function withoutStyle(): self
+    {
+        return new self($this->value, null, $this->comment);
+    }
+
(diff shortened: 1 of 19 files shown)


src/Common/Entity/Cell/BooleanCell.php
```diff
@@ -25,4 +25,34 @@ public function getValue(): bool
     {
         return $this->value;
     }
+
+    public static function create(bool $value): self
```
Slamdunk: Thank you very much for this PR, which indeed can be pretty helpful.
Just one thing: can we get rid of these `::create` API? What's the point of them if they are exactly the same as the constructor?

Slamdunk (changes requested):

src/Common/Entity/Cell/BooleanCell.php
karpilin: Thanks for almost immediate feedback.

I guess that's more of a personal habit at this point, but my intent was to treat `::create()` as the primary entry point, which would allow us to eventually make `__construct` private (or protected), which would give us flexibility to change internal instantiation logic later without breaking the public API.

However, if we want to keep the constructor public, I admit ::create is currently just syntactic sugar. let me know how you want me to go about it and I'll update the PR

Slamdunk: I'm not planning a new major version any time soon, so I'd rather merge only what really adds value. 

Please let's remove those create methods, and then we can ship this 💪

karpilin: all done and ready to ship :rocket:

Slamdunk (approved):

PR #342, opened 2025-12-11
Add Row::fromValuesWithStyle helper

### LINKED ISSUE #339: Issue: Unable to Apply Row-Level Style in Latest Version — Row::fromValues() No Longer Accepts Style Parameter
Hi @Slamdunk,

Previously, we were able to apply a row-level style by passing a Style object as the second parameter to Row::fromValues(), like this:

```
$titleStyle->setFontBold();
$titleStyle->setFontSize($this->fontSize);
$titleStyle->setFontColor(Color::toARGB("#" . $this->fontColor));
$titleStyle->setCellAlignment(CellAlignment::CENTER);

// Write data
$writer->addRow(Row::fromValues([$filterDetails], $titleStyle));
```

In the latest version (a27ee598ac09...), the Style parameter has been removed from Row::fromValues() and now only the $height parameter is accepted.

This raises a question:

### How should styles be applied in the new version?

- Is there a new recommended way to set a row-level style?

- Should we now apply styles per-cell instead of per-row?

- If row-level styling has been intentionally removed, is there an alternative pattern we should follow?

It would be helpful if the upgrade notes or documentation could clarify the new approach for applying styles after this change.

Thanks!

arokettu:
Addresses #339

Allows users to apply a single style to all cells in the row without creating unnecessary arrays

```php
// 5.0:

// we need to create a variable even if a row is hardcoded
$cellValues = ['column 1', 'column 2', 'column 3'];
// we need to count items and create an array
$columnStyles = array_fill(0, count($cellValues), new Style());

$writer->addRow(Row::fromValuesWithStyles($cellValues, $columnStyles));

// MR:
$writer->addRow(Row::fromValuesWithStyle(['column 1', 'column 2', 'column 3'], new Style()));
```

Adds a new method instead of reintroducing `$rowStyle` to `fromValues` to indicate its semantic difference

### DIFF
--- a/src/Common/Entity/Row.php
+++ b/src/Common/Entity/Row.php
@@ -103,4 +103,16 @@ final readonly class Row
 
         return new self($cells, $height);
     }
+
+    /**
+     * @param array<non-negative-int, null|bool|DateInterval|DateTimeInterface|float|int|string> $cellValues
+     */
+    public static function fromValuesWithStyle(array $cellValues, Style $cellStyle, float $height = 0): self
+    {
+        $cells = array_map(static function (bool|DateInterval|DateTimeInterface|float|int|string|null $cellValue) use ($cellStyle): Cell {
+            return Cell::fromValue($cellValue, $cellStyle);
+        }, $cellValues);
+
+        return new self($cells, $height);
+    }
 }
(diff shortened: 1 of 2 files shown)


arokettu:
not sure if I tested the right thing though

Slamdunk (approved):

PR #351, opened 2025-12-19
Add `with`ers to `Cell` abstract class

### LINKED ISSUE #350: Allow changing cell style
I like the immutable approach for the new styles since version 5, but I am lacking the ability to set the cell styles.

My use case: I'm gathering cells into an array before I call "addRow()". But before, I set the style of the last cell (it needs a final border).

Because many columns are optional, I don't know which will be the last cell until I'm at the end of the list.

The old approach was this:

```php
end($cells)->setStyle($styleLast);
$writer->addRow(new Row($cells));
```

I'm okay with replacing the cell with the new style, but for that a cloning mechanism is missing. So e.g.:

```php
$lastCell = array_pop($cells);
$cells[] = $lastCell->withStyle($styleLast);
$writer->addRow(new Row($cells));
```

Or simply allow to update the style of the cell. It shouldn't cause any problems here because it's not even given to the writer, yet.

uncaught:
Adding `withStyle` to Cell base class, closing #350

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "infection/infection": "^0.31.9",
         "phpbench/phpbench": "^1.4.3",
         "phpstan/phpstan": "^2.1.33",
-        "phpstan/phpstan-phpunit": "^2.0.10",
+        "phpstan/phpstan-phpunit": "^2.0.11",
         "phpstan/phpstan-strict-rules": "^2.0.7",
         "phpunit/phpunit": "^12.5.4"
     },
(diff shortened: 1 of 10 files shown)


uncaught:
@Slamdunk thanks for merging! Could you clarify, why `static` is not okay to use? It's available since PHP 8.0 and this library is ≥8.3

Slamdunk:
My bad sorry, fixed in https://github.com/openspout/openspout/pull/353

PR #355, opened 2026-01-19
Prevent OpenSpout export errors for cell values over 32,767 characters

### LINKED ISSUE #356: Export failure when a cell value exceeds Excel’s 32,767 character limit
Hi @Slamdunk,

🐛 **Problem**

When exporting spreadsheets using OpenSpout, if a cell value exceeds Excel’s maximum allowed character length of 32,767 characters, the library throws an exception during write operations. This behavior differs from **PhpOffice**, which internally handled or truncated values beyond this limit without error—so users of PhpOffice didn’t encounter this issue during exports.

📌 **Expected Behavior**

OpenSpout should either:
- gracefully handle values longer than Excel’s cell limit, or
- provide a clear documented recommendation (e.g., truncate, skip, warn) to avoid unexpected exceptions during export.

Excel itself imposes a hard limit of 32,767 characters per cell (Excel specification).

💡 **Proposed Solution**

In PR #355, I implemented logic to trim the cell value to 32,767 characters before writing it, preventing the export error and keeping the behavior aligned with Excel’s constraints. The trimming happens only when necessary, avoiding runtime exceptions for large text values.

✅ **Benefits**
- prevents uncaught exceptions when exporting large text values
- aligns output with Excel’s documented limits
- improves stability for applications migrating from PhpOffice to OpenSpout

Thanks!

nimitbhagat:
Hi @Slamdunk, 

I’ve raised this PR to address an issue encountered after switching from PhpOffice to OpenSpout.

Previously, PhpOffice handled cases where a cell’s content exceeded Excel’s 32,767 character limit by internally managing or truncating the data. After migrating to OpenSpout, the same scenario now results in an exception being thrown when the character limit is exceeded.

To ensure compatibility with Excel constraints and maintain stability, this PR proposes trimming any extra characters beyond the 32,767 limit before writing the value to the cell.

This change aligns the behavior with Excel’s specifications and avoids runtime errors while exporting large datasets.

Please let me know if you’d prefer a different approach or additional safeguards.

Thank You

### DIFF
--- a/src/Writer/XLSX/Manager/WorksheetManager.php
+++ b/src/Writer/XLSX/Manager/WorksheetManager.php
@@ -161,7 +161,7 @@ final readonly class WorksheetManager implements WorksheetManagerInterface
     private function getCellXMLFragmentForNonEmptyString(string $cellValue): string
     {
         if ($this->stringHelper->getStringLength($cellValue) > self::MAX_CHARACTERS_PER_CELL) {
-            throw new InvalidArgumentException('Trying to add a value that exceeds the maximum number of characters allowed in a cell (32,767)');
+            $cellValue = mb_substr($cellValue, 0, self::MAX_CHARACTERS_PER_CELL, 'UTF-8');
         }
 
         if ($this->options->SHOULD_USE_INLINE_STRINGS) {


Slamdunk:
> PhpOffice handled cases where a cell’s content exceeded Excel’s 32,767

> To ensure compatibility with Excel

Hello, I'm all for making this library as compatible with Excel as possible, but I don't care about being compatible with PhpOffice.

So the question is: what does Excel to when you enter more than 32,767 chars in a cell? Does it just truncate the data without notice, or does it pop a message up that inhibits the user from doing so?

nimitbhagat:
> > PhpOffice handled cases where a cell’s content exceeded Excel’s 32,767
> 
> > To ensure compatibility with Excel
> 
> Hello, I'm all for making this library as compatible with Excel as possible, but I don't care about being compatible with PhpOffice.
> 
> So the question is: what does Excel to when you enter more than 32,767 chars in a cell? Does it just truncate the data without notice, or does it pop a message up that inhibits the user from doing so?

Hello, 

No — Excel does not show any warning or popup when a user enters more than 32,767 characters in a cell.

Excel silently truncates the content to the 32,767-character limit without notifying the user. There is no validation error or UI feedback indicating that data has been cut off.

PR #365, opened 2026-03-09
fix: ignore row `spans` attribute when initializing row cells if `numColumns` contains value

### LINKED ISSUE #207: Inconsistent number of cells per row
If I read [Excel_template_500Users.xlsx](https://github.com/openspout/openspout/files/13660857/Excel_template_500Users.xlsx), rows until row 16 (Uer15) have consistent dimension (20 cells), after that (User16) it's reduced to 5 cells

aprat84:
Tries to solve issue #207.

This is not a perfect solution, cause we are depending on the correct value inside `<dimension>` tag. But the only alternative is to parse the entire sheet first!

I did also associate some related tests with their appropiate fixture XLSX, according to test and XLSX file names.

_PD: naming suggestions accepted (code and commit message 😅)_


tests/Reader/XLSX/ReaderTest.php
```diff
@@ -421,17 +421,34 @@ public function testReadShouldKeepEmptyCellsAtTheEndIfDimensionsSpecified(): voi
...
+            ['s1--A1', 's1--B1', 's1--C1', 's1--D1', 's1--E1'],
+            ['s1--A2', 's1--B2', 's1--C2', '', ''],
+        ];
+        self::assertSame($expectedRows, $allRows);
+    }
+
     public function testReadShouldKeepEmptyCellsAtTheEndIfNoDimensionsButSpansSpecified(): void
     {
-        $allRows = $this->getAllRowsForFile('sheet_without_dimensions_and_empty_cells.xlsx');
+        $allRows = $this->getAllRowsForFile('sheet_without_dimensions_but_spans_and_empty_cells.xlsx');
```
Slamdunk: This is confusing: by swapping the content of two tests you got the `git-history` dirty.
Please revert the previously existing tests and only edit the expected values where necessary

Slamdunk (changes requested):

tests/Reader/XLSX/ReaderTest.php
aprat84: Are you sure? I just matched the test name with the XLSX file name that uses.

The edit of the `foreach` below is to assert according to the new behavior.

The test for when there's no dimensions or spans is `testReadShouldSkipEmptyCellsAtTheEndIfDimensionsNotSpecified`.

IMHO there should be a test for every combination of:
* existence of dimensions tag
* existence of spans attribute

Which is 4 tests, as there are. Maybe we could improve the naming...

Slamdunk: Then change the test names please.

Understanding a change of a test content is hard.
Tracking a name change is easy.

Slamdunk (approved):

tests/Reader/XLSX/ReaderTest.php
aprat84: I did the best I could!

From lines 408 to 486 in `tests/Reader/XLSX/ReaderTest.php`, there are 4 tests, including the new test added by this PR.

As I said before, there are 4 combinations to cover:
| # | Dimensions tag | Spans attribute |
|--------|--------|--------|
| 1 | ❌ | ❌ |
| 2 | ❌ | ✅ |
| 3 | ✅ | ❌ |
| 4 | ✅ | ✅ |


The test I added for this PR covers the case number 4.

Test in line 456 covers the case number 1. 
* Name `testReadShouldSkipEmptyCellsAtTheEndIfDimensionsNotSpecified`
* No changes needed.

Test in line 408 covers case number 2.
* Name changed to `testReadShouldKeepEmptyCellsAtTheEndIfNoDimensionsButSpansSpecified`

Test in line 440 needs to cover case number 3, but it is a copy of the test in line 456, so it needs full refactor:
* Name changed to `testReadShouldKeepEmptyCellsAtTheEndIfDimensionsSpecified`
* Fixture file changed to `sheet_with_dimensions_and_empty_cells.xlsx`
* Asserts changed according to how the library works: if there's a dimensions tag, it honours it and empty cells are created.


If this is not enough, I'm afraid I'll leave this PR with only the first commit...

Slamdunk: Your PR has already been released, thank you :wink:

uncaught: @Slamdunk with this version most of our tests fail to match their previous snapshot. They all seem to add new empty cells at the end of rows. Is that intentional?

Slamdunk: Yes it is intentional, I'm sorry if this represents a BC break in your case

aprat84: @uncaught could you provide an XLSX? Just to investigate.

Tip: unzip the XLSX, remove sensible data in the XMLs, and zip it again as XLSX.

uncaught: @Slamdunk it's fine, I've simply updated the snapshots. It was just unexpected thus the question.

@aprat84 it's not failing, it is simply producing slightly different results now.

Before:

```
[
  ['r1c1', 'r1c2'],
  ['r2c1', 'r2c2', 'r2c3'],
]
```

Now:

```
[
  ['r1c1', 'r1c2', ''],
  ['r2c1', 'r2c2', 'r2c3'],
]
```

It's just padding rows with empty cells, that's all. I'm fine with that, just wanted to make sure it was intentional 👍

aprat84: @uncaught Yeah, bad word selection, didn't mean "failing" strictly.

Just wondered which combination of dimensions tag and spans attribute do your XLSX have!

I'll assume that both are specified, which is the case this PR tried to solve, and be more consistent. Now the only case where no "padded" rows with empty cells is when there is no dimensions tag and no spans attribute, which I believe its a very rare case!

uncaught: Then I think I must be a rare case ;). I'm generating Excel files with this library and then re-read them in the unit tests. And some header/footer or spacer rows don't get fully filled. E.g. something like:

```
['Project:', 'My project'],
['Time period:', 'October 2025'],
[],
['Date', 'Time', 'Comment'],
['19.03.2026', '07:45:22', 'I want to be a unicorn'],
['19.03.2026', '07:45:22', 'I want to be a unicorn'],
['19.03.2026', '07:45:22', 'I want to be a unicorn'],
...
```

PR #368, opened 2026-03-14
feat: add FIELD_ESCAPE option for CSV reader and EOL option for CSV w…

### LINKED ISSUE #366: Support customizing $eol and $escape in CSV writer via Options
The CSV writer currently hardcodes the $escape ('') and $eol ("\n") parameters when calling fputcsv():

```php
fputcsv(
    $this->filePointer,
    $cells,
    $this->options->FIELD_DELIMITER,
    $this->options->FIELD_ENCLOSURE,
    ''
);
```

PHP 8.1 added an $eol parameter to fputcsv(), making it possible to control the line ending per row. Currently there is no way to configure either $escape or $eol through Options.

This matters because https://www.rfc-editor.org/rfc/rfc4180 requires CRLF (\r\n) as the record delimiter. The current output uses bare LF (\n), which means CSV files produced by OpenSpout are not strictly RFC 4180 compliant.

**Proposal:**
Add two new properties to OpenSpout\Writer\CSV\Options:
```php
public string $FIELD_ESCAPE = '';
public string $EOL = "\n";
```

And pass them through in the writer:
```php
fputcsv(
    $this->filePointer,
    $cells,
    $this->options->FIELD_DELIMITER,
    $this->options->FIELD_ENCLOSURE,
    $this->options->FIELD_ESCAPE,
    $this->options->EOL,
);
```

This would keep the current default behavior unchanged while letting users opt in to RFC 4180 compliant output with:
```php
$options = new Options();
$options->EOL = "\r\n";
```

The $eol parameter requires PHP 8.1+, which is already the minimum version for OpenSpout 4.x.

Would you be open to a PR for this?

BoyeMagnus:
## Summary
- Add configurable `FIELD_ESCAPE` to `\OpenSpout\Reader\CSV\Options`, passed through to `fgetcsv()` (default `''` for backwards compatibility)
- Add configurable `EOL` to `\OpenSpout\Writer\CSV\Options`, passed through to `fputcsv()` (default `"\n"` for backwards compatibility), enabling RFC 4180 compliant CRLF output
- `FIELD_ESCAPE` is intentionally not exposed on the writer because PHP's `fputcsv()` always escapes the enclosure by doubling it, regardless of the escape parameter

## Test plan
- [x] New test `testReadShouldSupportCustomFieldEscape` with dedicated test resource
- [x] New tests `testWriteShouldSupportCrlfEol` and `testWriteShouldDefaultToLfEol`
- [x] All existing CSV reader and writer tests pass unchanged

See #366 for reference

### DIFF
--- a/src/Reader/CSV/Options.php
+++ b/src/Reader/CSV/Options.php
@@ -12,6 +12,7 @@ final readonly class Options
         public bool $SHOULD_PRESERVE_EMPTY_ROWS = false,
         public string $FIELD_DELIMITER = ',',
         public string $FIELD_ENCLOSURE = '"',
+        public string $FIELD_ESCAPE = '',
         public string $ENCODING = EncodingHelper::ENCODING_UTF8,
     ) {}
 }
--- a/src/Reader/CSV/Options.php
+++ b/src/Reader/CSV/Options.php
@@ -12,7 +12,7 @@ final readonly class Options
         public bool $SHOULD_PRESERVE_EMPTY_ROWS = false,
         public string $FIELD_DELIMITER = ',',
         public string $FIELD_ENCLOSURE = '"',
-        public string $FIELD_ESCAPE = '',
         public string $ENCODING = EncodingHelper::ENCODING_UTF8,
+        public string $FIELD_ESCAPE = '',
     ) {}
 }
--- a/src/Writer/CSV/Options.php
+++ b/src/Writer/CSV/Options.php
@@ -18,8 +18,8 @@ final readonly class Options
     public function __construct(
         public string $FIELD_DELIMITER = ',',
         public string $FIELD_ENCLOSURE = '"',
-        public string $EOL = "\n",
         public bool $SHOULD_ADD_BOM = true,
         public int $FLUSH_THRESHOLD = 500,
+        public string $EOL = "\n",
     ) {}
 }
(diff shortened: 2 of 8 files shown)


src/Reader/CSV/Options.php
```diff
@@ -12,6 +12,7 @@ public function __construct(
         public bool $SHOULD_PRESERVE_EMPTY_ROWS = false,
         public string $FIELD_DELIMITER = ',',
         public string $FIELD_ENCLOSURE = '"',
+        public string $FIELD_ESCAPE = '',
```
Slamdunk: Hi, this is a breaking change, new parameters have to be appended in the argument order

src/Writer/CSV/Options.php
```diff
@@ -7,13 +7,18 @@
...
+     * escape parameter. FIELD_ESCAPE is only available on the reader Options.
+     *
      * @param non-empty-string $FIELD_DELIMITER
      * @param non-empty-string $FIELD_ENCLOSURE
      * @param positive-int     $FLUSH_THRESHOLD
      */
     public function __construct(
         public string $FIELD_DELIMITER = ',',
         public string $FIELD_ENCLOSURE = '"',
+        public string $EOL = "\n",
```
Slamdunk: Same here

Slamdunk (changes requested):
Few tweaks and we can ship this 💪

src/Reader/CSV/Options.php
BoyeMagnus: Ahh sorry of course, let me correct it straight away

Slamdunk (approved):

PR #369, opened 2026-03-16
XLSX: add support for Data Validation

### LINKED ISSUE #364: Possible Feature: native data validation (dropdown lists) support for XLSX
Hi! I've been using OpenSpout for generating XLSX exports and found myself needing dropdown validation on certain columns. I ended up implementing a post-processing step that directly manipulates the ZIP/XML after `$writer->close()`, but it feels like something that could fit naturally into the library.

Is this something you'd want to include, or would it complicate the library too much? 

If you are interested, is it okay if I make a pull request for this?

Is it okay if the PR would only cover `type="list"` (dropdown options), or would you want it to support the other `<dataValidation>` types too, such as whole number ranges, decimal ranges, date ranges, and text length constraints? I'm happy to start with just the list type and leave the rest for follow-up PRs if you'd prefer to keep the scope small.

## Here is a sketch of how I would implement it:

I would add another methode inside `createContentFiles()` in `FileSystemHelper`, which is where all the worksheet XML is assembled sequentially. 

Per the OOXML spec, `<dataValidations>` belongs after `<mergeCells>` and before `<pageMargins>`, so it slots in right after the existing merge cells block.
The implementation would mirror the MergeCell pattern exactly:

1. A `DataValidation` value object in `src/Writer/XLSX/` at the same level as `MergeCell.php`
2. `Options::addDataValidation()` and `Options::getDataValidations()` following the same signature style as `Options::mergeCells()` and `Options::getMergeCells()`
3. ~15 lines added to `createContentFiles()` in FileSystemHelper, right after the merge cells block

```php
// create nodes for data validations
$dataValidations = array_filter(
    $options->getDataValidations(),
    static fn (DataValidation $dv) => $dv->sheetIndex === $worksheet->getExternalSheet()->getIndex(),
);
if ([] !== $dataValidations) {
    $dataValidationString = '<dataValidations count="'.\count($dataValidations).'">';
    foreach ($dataValidations as $dataValidation) {
        $optionList = implode(',', $dataValidation->options);
        $dataValidationString .= \sprintf(
            '<dataValidation type="list" allowBlank="1" showInputMessage="1" showErrorMessage="1" sqref="%s"><formula1>&quot;%s&quot;</formula1></dataValidation>', // This could be made configurable as well
            htmlspecialchars($dataValidation->sqref, \ENT_XML1),
            $optionList,
        );
    }
    $dataValidationString .= '</dataValidations>';
    fwrite($worksheetFilePointer, $dataValidationString);
}
```

The proposed API would look like this:

```php
php$options = new Options();
$writer->openToFile('/tmp/file.xlsx');

$writer->addRow(Row::fromValues(['Name', 'Status']));

$options->addDataValidation(
    sheetIndex: 0,
    sqref: 'B2:B100',
    options: ['Active', 'Inactive', 'Pending'],
);

$writer->close();
```

One known limitation worth flagging: the inline list format ("Foo,Bar,Baz") uses commas as delimiters with no escape mechanism in the OOXML spec, so option values containing literal commas cannot be represented faithfully. This is a spec constraint, not an implementation one.

michieljmmaas:
Base on this issue: https://github.com/openspout/openspout/issues/364

## Add data validation support for XLSX

This PR adds support for all data validation types:

- Whole number
- Decimal
- Date
- Time
- Text length
- List (inline options and cell range reference)
- Custom formula

Configured via `Options::addValidation()`, following the same pattern as `mergeCells`.

## Usage
```php
$options = new Options();
$writer->openToFile('/tmp/file.xlsx');

$writer->addRow(Row::fromValues(['Name', 'Status', 'Score', 'Date']));

$options->addValidation(
    sheetIndex: $writer->getCurrentSheet()->getIndex(),
    topLeftColumn: 2,
    topLeftRow: 2,
    bottomRightColumn: 2,
    bottomRightRow: 100,
    rule: new WholeNumberValidationRule(ValidationOperator::Between, 0, 100),
    validation_display: new ValidationDisplay(
        allowBlank: false,
        showInputMessage: true,
        showErrorMessage: true,
        errorStyle: ErrorStyle::Stop,
        promptTitle: 'Score',
        prompt: 'Enter a score between 0 and 100',
        errorTitle: 'Invalid score',
        error: 'Score must be a whole number between 0 and 100',
    ),
);

$writer->close();
```

### Support for Raw Values and CellReferences

I added support for both Raw PHP Values when selection what you want to compare. This is different from what [PHPSpreadsheet](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Cell/DataValidation.php) and [openpyxl ](https://github.com/soxhub/openpyxl/blob/master/openpyxl/worksheet/datavalidation.py) do, where the only allow string inputs. It felt this was a not that nice to work with. 

Next to the typed values you can also provide a `CellReference`, this is the default way Excel uses it. 

### Throwing exceptions on bad input

Constructors throw `\InvalidArgumentException` for badly formatted data. Like:

- A comma in an `ListValidationRule` option (commas are the delimiter, there's no way to escape them)
- An empty options list in `ListValidationRule`
- Inverted coordinates in `CellReference`
- `TimeOfDay` guarding for correct time
- `between`/`notBetween` used with only one value

`MergeCell` doesn't do any of this, so it's a bit inconsistent with the rest of the codebase. Happy to remove the guards if you think they are too much. I would like to keep the `between`/`notBetween` guard, as removing it could generate invalid XML. 

### ODS support
I looked into ODS support for validations, but chose not to build it since there was no need for it. 
In theory the current validation DTO and `ValidationDisplay` logic could be shared, and be thus be placed in a `Common` directory. I chose not to do this because placing it there, without a ODS implement is a bit confusing. 

If we were to implement it later, and reuse those classes, it would mean moving them, and causing a breaking change for the people already using them. 

Would do you think about this?


michieljmmaas:
I fixed the failing workflows (PHPStan and php-cs), fixed the commits and force pushed my changes.

src/Writer/XLSX/Validation/Rules/ListValidationRule.php
```diff
@@ -0,0 +1,41 @@
...
+                if (str_contains($option, ',')) {
+                    throw new InvalidArgumentException(
+                        \sprintf(
+                            "List validation option '%s' contains a comma, which is not allowed as it is used as a delimiter.",
+                            $option,
+                        )
+                    );
+                }
+            }
+        }
```
Slamdunk: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

```suggestion
        if ($this->value instanceof CellReference) {
            return;
        }
		if ([] === $this->value) {
			throw new InvalidArgumentException('ListValidationRule requires at least one option.');
		}
		foreach ($this->value as $option) {
			if (! str_contains($option, ',')) {
			    continue;
			}
			throw new InvalidArgumentException(
				\sprintf(
					"List validation option '%s' contains a comma, which is not allowed as it is used as a delimiter.",
					$option,
				)
			);
		}
```

src/Writer/XLSX/Validation/Rules/ListValidationRule.php
```diff
@@ -0,0 +1,41 @@
...
+
+use InvalidArgumentException;
+use OpenSpout\Writer\XLSX\Validation\CellReference;
+use OpenSpout\Writer\XLSX\Validation\DataValidationRuleInterface;
+use OpenSpout\Writer\XLSX\Validation\ValidationRuleType;
+
+final readonly class ListValidationRule implements DataValidationRuleInterface
+{
+    /**
+     * @param CellReference|list<string> $value
```
Slamdunk:
```suggestion
     * @param CellReference|non-empty-list<non-empty-string> $value
```

src/Writer/XLSX/Validation/Rules/TimeOfDay.php
```diff
@@ -0,0 +1,37 @@
...
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\XLSX\Validation\Rules;
+
+use InvalidArgumentException;
+
+final readonly class TimeOfDay
+{
+    public function __construct(
```
Slamdunk, line 14:
```suggestion
    /**
     * @param int<0, 23> $hours
     * @param int<0, 59> $minutes
     * @param int<0, 59> $seconds
     */
    public function __construct(
```

src/Writer/XLSX/Validation/RuleSerializer/ValidationRuleSerializer.php
```diff
@@ -0,0 +1,183 @@
...
+use OpenSpout\Writer\XLSX\Validation\Rules\CustomValidationRule;
+use OpenSpout\Writer\XLSX\Validation\Rules\DateValidationRule;
+use OpenSpout\Writer\XLSX\Validation\Rules\DecimalValidationRule;
+use OpenSpout\Writer\XLSX\Validation\Rules\ListValidationRule;
+use OpenSpout\Writer\XLSX\Validation\Rules\TextLengthValidationRule;
+use OpenSpout\Writer\XLSX\Validation\Rules\TimeOfDay;
+use OpenSpout\Writer\XLSX\Validation\Rules\TimeValidationRule;
+use OpenSpout\Writer\XLSX\Validation\Rules\WholeNumberValidationRule;
+
+final class ValidationRuleSerializer
```
Slamdunk: I don't like much the purpose of this class: since each validation rule needs to be serialized anyway, can we drop this and add a `serialize()` API on `DataValidationRuleInterface`, and let each validation rule do its own job?

src/Writer/XLSX/Validation/CellReference.php
```diff
@@ -0,0 +1,27 @@
...
+
+use InvalidArgumentException;
+
+final readonly class CellReference
+{
+    /**
+     * @param int<0, max> $topLeftColumn
+     * @param int<0, max> $topLeftRow
+     * @param int<0, max> $bottomRightColumn
+     * @param int<0, max> $bottomRightRow
```
Slamdunk:
```suggestion
     * @param non-negative-int $topLeftColumn
     * @param non-negative-int $topLeftRow
     * @param non-negative-int $bottomRightColumn
     * @param non-negative-int $bottomRightRow
```

src/Writer/XLSX/Validation/ValidationDisplay.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\XLSX\Validation;
+
+final class ValidationDisplay
```
Slamdunk:
```suggestion
final readonly class ValidationDisplay
```

src/Writer/XLSX/Validation/ValidationDisplay.php
```diff
@@ -0,0 +1,24 @@
...
+        public bool $showInputMessage = true,
+        public bool $showErrorMessage = true,
+        public ErrorStyle $errorStyle = ErrorStyle::Stop,
+        public ?string $promptTitle = null,
+        public ?string $prompt = null,
+        public ?string $errorTitle = null,
+        public ?string $error = null,
+    ) {}
+
+    public static function default(): self
```
Slamdunk: I don't see this method used anywhere, can we drop it please?

src/Writer/XLSX/Validation/ValidationRule.php
```diff
@@ -0,0 +1,25 @@
...
+namespace OpenSpout\Writer\XLSX\Validation;
+
+final readonly class ValidationRule
+{
+    /**
+     * @param 0|positive-int $sheetIndex
+     * @param 0|positive-int $topLeftColumn
+     * @param positive-int   $topLeftRow
+     * @param 0|positive-int $bottomRightColumn
+     * @param positive-int   $bottomRightRow
```
Slamdunk:
```suggestion
     * @param non-negative-int $sheetIndex
     * @param non-negative-int $topLeftColumn
     * @param positive-int   $topLeftRow
     * @param non-negative-int $bottomRightColumn
     * @param positive-int   $bottomRightRow
```

Slamdunk (changes requested):
Awesome, few tweaks and we can ship it 💪

src/Writer/XLSX/Validation/ValidationDisplay.php
michieljmmaas: Sure. This isn't really needed.

src/Writer/XLSX/Validation/RuleSerializer/ValidationRuleSerializer.php
michieljmmaas: This is how I had it initially, but is changed my opinion.

Keeping the `DataValidationRuleInterface`  as pure DTO's means they could eventually be reused for a ODS implementation. If we add that method to the interface, and implement it on all the current rules, the implemenation would no longer be platform agnostic. If we want to add a DTO implementation later down the line it would mean that all this logic would have to be duplicate ODS implementation. Using a serialize pattern like this would only mean that we would need to a new serializers for the DSO implementation. 

Also adding a method like that makes these Data Object as 'pure', which I would prefer. 

I do however respect your opinion as a maintainer. If you would prefer to add a serialize method to it, I can implement it for you, no problem!

I would do it like this:
- Add `serialize()` to the interface and implement it.
- Add a `serialize()` method to the `CellReference` 
- Remove the `ValidationRuleSerializer`  class
- Move the tests in the `ValidationRuleSerializerTest` to individual test classes for all the rules, and remove that the test as well.

michieljmmaas:
@Slamdunk Thank you for the fast reply and thorough review! 💪 

I accepted all you suggestions, and left a reply on my thoughts on the current serializer-pattern. 

After the review is done, do you want me to squash/fix-up the commits to make the changes easier to read?

src/Writer/XLSX/Validation/RuleSerializer/ValidationRuleSerializer.php
Slamdunk: > Keeping the `DataValidationRuleInterface` as pure DTO's means they could eventually be reused for a ODS implementation.

This is a noble and smart target.

Fact is I don't care anymore: ODS is too different from XLSX, no one has ever needed to write both at the same time, so our effort to maintain interoperability at language level is just wasted time.

michieljmmaas:
Okay, I moved the serialization logic from the static class to the validation rules themselves.

I also updated the `CellReference`  to be more like the `AutoFilter` class, for some better consistency.

Let me know if you agree with these changes.

If you want I can also recreate the commits, to make the git history more legible

Slamdunk (approved):

Issue #172, opened 2023-08-29
insert image to cell or row - Request


eabbas:
inserting images can give more power to this package if it is possible.

thelfensdrfer:
Still a good idea!

Slamdunk:
Any PR is welcome

thelfensdrfer:
> Any PR is welcome

Sure, just wanted to prevent the github bot from closing the issue.

thelfensdrfer:
PR is open, so please don't close it for now.

PR #378, opened 2026-04-24
Xlsx images

thelfensdrfer:
Regarding https://github.com/openspout/openspout/issues/172, it was easier than I initially thought to implement, so I gave it a try.

## Caveats
- I don't know anything about XLSX; this part is AI-generated.
- Only works for XLSX files for now; (of course, CSV will never be an option.)
- Image is inside the cell (either fitting or with full dimensions) and not absolutely positioned

## Example
```php
<?php

require_once "vendor/autoload.php";

use OpenSpout\Common\Entity\Cell;

// Analyze image
$path = 'test.png';
$imagick = new Imagick($path);
$width = $imagick->getImageWidth();
$height = $imagick->getImageHeight();
$imagick->clear();

// Create xlsx
$writer = new Writer;
$writer->openToFile(__DIR__ . '/test.xlsx');

$cells = [
    new Cell\ImageCell(
        path: $path,
        width: $width,
        height: $height,
        fitToCell: true,
    ),
    Cell::fromValue('Carl'),
    Cell::fromValue('is'),
    Cell::fromValue('great!'),
];

$singleRow = new Row($cells);
$writer->addRow($singleRow);

$writer->close();
```

### DIFF
--- a/src/Common/Entity/Cell.php
+++ b/src/Common/Entity/Cell.php
@@ -73,7 +73,7 @@ abstract readonly class Cell
         return new StringCell($value, $style, $comment, $hyperlinkUrl);
     }
 
-    final public static function fromImageFile(
+    final public static function fromImagePath(
         string $path,
         int $width,
         int $height,
--- a/src/Common/Entity/Cell.php
+++ b/src/Common/Entity/Cell.php
@@ -72,15 +72,4 @@ abstract readonly class Cell
 
         return new StringCell($value, $style, $comment, $hyperlinkUrl);
     }
-
-    final public static function fromImagePath(
-        string $path,
-        int $width,
-        int $height,
-        ?Style $style = null,
-        ?Comment $comment = null,
-        bool $fitToCell = false,
-    ): self {
-        return new ImageCell($path, $width, $height, $style, $comment, $fitToCell);
-    }
 }
--- a/src/Common/Entity/Cell.php
+++ b/src/Common/Entity/Cell.php
@@ -11,6 +11,7 @@ use OpenSpout\Common\Entity\Cell\DateIntervalCell;
 use OpenSpout\Common\Entity\Cell\DateTimeCell;
 use OpenSpout\Common\Entity\Cell\EmptyCell;
 use OpenSpout\Common\Entity\Cell\FormulaCell;
+use OpenSpout\Common\Entity\Cell\ImageCell;
 use OpenSpout\Common\Entity\Cell\NumericCell;
 use OpenSpout\Common\Entity\Cell\StringCell;
 use OpenSpout\Common\Entity\Cell\TextRunCell;
@@ -69,5 +70,14 @@ abstract readonly class Cell
         }
 
         return new StringCell($value, $style, $comment);
+    }
+
+    final public static function fromFile(
+        string $path,
+        ?Style $style = null,
+        ?Comment $comment = null,
+        bool $fitToCell = false,
+    ): self {
+        return new ImageCell($path, $style, $comment, $fitToCell);
     }
 }
--- a/src/Common/Entity/Cell.php
+++ b/src/Common/Entity/Cell.php
@@ -75,10 +75,12 @@ abstract readonly class Cell
 
     final public static function fromImageFile(
         string $path,
+        int $width,
+        int $height,
         ?Style $style = null,
         ?Comment $comment = null,
         bool $fitToCell = false,
     ): self {
-        return new ImageCell($path, $style, $comment, $fitToCell);
+        return new ImageCell($path, $width, $height, $style, $comment, $fitToCell);
     }
 }
--- a/src/Common/Entity/Cell.php
+++ b/src/Common/Entity/Cell.php
@@ -11,7 +11,6 @@ use OpenSpout\Common\Entity\Cell\DateIntervalCell;
 use OpenSpout\Common\Entity\Cell\DateTimeCell;
 use OpenSpout\Common\Entity\Cell\EmptyCell;
 use OpenSpout\Common\Entity\Cell\FormulaCell;
-use OpenSpout\Common\Entity\Cell\ImageCell;
 use OpenSpout\Common\Entity\Cell\NumericCell;
 use OpenSpout\Common\Entity\Cell\StringCell;
 use OpenSpout\Common\Entity\Cell\TextRunCell;
--- a/src/Common/Entity/Cell.php
+++ b/src/Common/Entity/Cell.php
@@ -73,7 +73,7 @@ abstract readonly class Cell
         return new StringCell($value, $style, $comment, $hyperlinkUrl);
     }
 
-    final public static function fromFile(
+    final public static function fromImageFile(
         string $path,
         ?Style $style = null,
         ?Comment $comment = null,
--- a/src/Writer/Common/Entity/Worksheet.php
+++ b/src/Writer/Common/Entity/Worksheet.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace OpenSpout\Writer\Common\Entity;
 
+use OpenSpout\Common\Entity\Cell\ImageCell;
+
 /**
  * Entity describing a Worksheet.
  */
@@ -23,6 +25,9 @@ final class Worksheet
 
     /** @var int Index of the last written row */
     private int $lastWrittenRowIndex = 0;
+
+    /** @var array<array{row: int, col: int, cell: ImageCell}> */
+    private array $images = [];
 
     /**
      * Worksheet constructor.
@@ -79,6 +84,17 @@ final class Worksheet
     public function setLastWrittenRowIndex(int $lastWrittenRowIndex): void
     {
         $this->lastWrittenRowIndex = $lastWrittenRowIndex;
+    }
+
+    public function addImage(int $row, int $col, ImageCell $cell): void
+    {
+        $this->images[] = ['row' => $row, 'col' => $col, 'cell' => $cell];
+    }
+
+    /** @return array<array{row: int, col: int, cell: ImageCell}> */
+    public function getImages(): array
+    {
+        return $this->images;
     }
 
     /**
--- a/src/Writer/Common/Entity/Worksheet.php
+++ b/src/Writer/Common/Entity/Worksheet.php
@@ -91,7 +91,9 @@ final class Worksheet
         $this->images[] = ['row' => $row, 'col' => $col, 'cell' => $cell];
     }
 
-    /** @return array<array{row: int, col: int, cell: ImageCell}> */
+    /**
+     * @return array<array{row: int, col: int, cell: ImageCell}>
+     */
     public function getImages(): array
     {
         return $this->images;
--- a/src/Writer/Common/Entity/Worksheet.php
+++ b/src/Writer/Common/Entity/Worksheet.php
@@ -26,7 +26,7 @@ final class Worksheet
     /** @var int Index of the last written row */
     private int $lastWrittenRowIndex = 0;
 
-    /** @var array<array{row: int, col: int, cell: ImageCell}> */
+    /** @var list<array{row: int, col: int, cell: ImageCell}> */
     private array $images = [];
 
     /**
@@ -92,7 +92,7 @@ final class Worksheet
     }
 
     /**
-     * @return array<array{row: int, col: int, cell: ImageCell}>
+     * @return list<array{row: int, col: int, cell: ImageCell}>
      */
     public function getImages(): array
     {
--- a/tests/Writer/CSV/WriterTest.php
+++ b/tests/Writer/CSV/WriterTest.php
@@ -279,7 +279,7 @@ final class WriterTest extends TestCase
     {
         // Minimal 1×1 image
         $imagePath = sys_get_temp_dir().\DIRECTORY_SEPARATOR.'openspout_test_image_csv.png';
-        file_put_contents($imagePath, base64_decode('iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwADhQGAWjR9awAAAABJRU5ErkJggg=='));
+        file_put_contents($imagePath, base64_decode('iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwADhQGAWjR9awAAAABJRU5ErkJggg==', true));
 
         try {
             $fileName = 'test_image_cell_throws.csv';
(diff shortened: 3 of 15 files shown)


thelfensdrfer:
@Slamdunk maybe you can take a look.

I have a problem with php-cs-fixer. I do not want to import the Imagick class globally, in case it is not installed. This pr works with gd and imagick, so imagick is not required. I cannot get `@php-cs-fixer-ignore` to work, to ignore the global namespace rule in the 4 files which import Imagick.

Other than that, it works for me.

src/Common/Entity/Cell.php
```diff
@@ -71,4 +72,13 @@ final public static function fromValue(
 
         return new StringCell($value, $style, $comment, $hyperlinkUrl);
     }
+
+    final public static function fromFile(
```
michieljmmaas: Using `fromFile` here might be a bit misleading. I would suggest using something like `fromImageFile` or `fromImage`.

michieljmmaas (commented):
Great to see someone picking this up - I'd love to see this feature land.

I think you may have scared maintainers off a bit with your caveats. I took a look myself and was pretty happy with what I saw. (To be clear: I'm not a maintainer, just a [contributor](https://github.com/openspout/openspout/pull/369) who'd like this feature.)

## Scoping to just `ImageCell` seems like a good idea

Starting with `ImageCell` rather than a free-floating image seems like a sensible choice! It reuses the existing cell/row flow and fits the streaming model. Worth noting in the PR description that floating/absolute-anchored images are out of scope (or a follow-up), so reviewers know the boundaries.

## Image sizes

One area I'd love the maintainers' thoughts on: the hard dependency on GD or Imagick just to read image dimensions. The fallback logic adds a fair amount of complexity, and adding an extension requirement to a deliberately lean library feels like something Slamdunk might push back on. Would it be better to let the caller provide the `width` and `height` explicitly? That gives the implementation a way to resize images, pushes the responsibility to the user (who likely has those extensions enabled anyway), and lets you remove all the GD/Imagick references from the code entirely. The tests could use actual image files to cover the real-world case.

This would mean users have to know their image dimensions up front, but that feels in keeping with OpenSpout's lean philosophy - if you want images in your spreadsheet, you decide how they're sized.

If the maintainers prefer keeping the auto-detection path, it would be worth adding GD (or Imagick) to the project's `Dockerfile` so the new tests actually run in the dev environment - I had to add it locally before I could exercise the code.

## Results

To help with review, I built the PR locally and attached two files it produced - they open cleanly in Excel.

[fit-to-cell.xlsx](https://github.com/user-attachments/files/27645979/fit-to-cell.xlsx)
[red-image.xlsx](https://github.com/user-attachments/files/27645980/red-image.xlsx)

thelfensdrfer:
@michieljmmaas thanks for your review. The dimension part actually sounds good. That would also resolve my issue with php-cs-fixer...

That would also make the pr easier to review, I will provide an update in the next days.

thelfensdrfer:
I updated the branch, so that ImageMagick/GD is no longer needed. Instead, the user has to provide the image size themselves.

luanmm:
I think maybe there are some points to add to the discussion.

Scenario 1: multiple cells with same image (i.e. when exporting a general app/system report, like order items). Considering this, ideally the image should appear just one time in the XLSX archive, being referenced by multiple cells. In the current implementation, it seems that it will appear (as embedded files) as many times it would appear in cells. Proposal: an `ImageManager` to handle image references (a map between image source/hash and its incremental ID).

Scenario 2: different origins (i.e. file path, internet URL, etc). Proposal: specify the public API in a more future-proof manner, using a more generic naming like `fromUri`. The same `ImageManager` could be the general class to process the images (download or get information from them) and could avoid using GD/Imagick if the base information is provided, but can be considered when the information is missing (as PHP is somehow driven by a "it just works" idea). 

Scenario 3: partially supported features (i.e. Excel supports "image by reference"). There are some updated versions that loads images from internet, for instance, when it has just the source URL referenced in the XML files (in that case, you would not have an embedded image in the XLSX archive). Proposal: find a way to pass options to the library (aligned to the current library spec), so the `ImageManager` could know what to do (reference/embed images).

Obviously, these scenarios could be considered in the future, as it would be improvements to this PR. The idea is just to point out some possibilities to harden the spec, avoiding breaking changes in the future.

-----

Disclaimer: I'm not a maintainer, just a developer that would use image features too.

thelfensdrfer:
> Obviously, these scenarios could be considered in the future, as it would be improvements to this PR. The idea is just to point out some possibilities to harden the spec, avoiding breaking changes in the future.

Agreed, and if a maintainer would give their opinion on this matter, I would be willing to implement a future proof version like you suggested. But without implementing the options, this could really be done in the future. 

Thanks for your input!

Slamdunk:
I am no longer able to focus approprietly on such topic, but I can trust PR where you guys agreed on.
Please remember that this library craves for speed: if you need maximum elaxticity use https://github.com/phpoffice/phpspreadsheet instead, please.

thelfensdrfer:
> I am no longer able to focus approprietly on such topic, but I can trust PR where you guys agreed on.

Sounds good!


> Please remember that this library craves for speed: if you need maximum elaxticity use https://github.com/phpoffice/phpspreadsheet instead, please.

No problem, this PR only adds overhead when you add an image. It doesn't compromise existing performance metrics.

thelfensdrfer:
@luanmm took a little bit longer, but the code is updated. Is that how you meant it? I used a similar (almost identical) approach as the recently added HyperLinkManager.

luanmm:
@thelfensdrfer nice catch! Great contribution! Thank you for all the work/initiative!

thelfensdrfer:
Only because I need it in https://github.com/filamentphp/filament tbh 😄

src/Common/Entity/Cell.php
```diff
@@ -71,4 +72,15 @@ final public static function fromValue(
 
         return new StringCell($value, $style, $comment, $hyperlinkUrl);
     }
+
+    final public static function fromImagePath(
```
Slamdunk: What's the point of this method, that's never used anywhere?

src/Writer/Common/Entity/Worksheet.php
```diff
@@ -24,6 +26,9 @@ final class Worksheet
     /** @var int Index of the last written row */
     private int $lastWrittenRowIndex = 0;
 
+    /** @var array<array{row: int, col: int, cell: ImageCell}> */
```
Slamdunk:
```suggestion
    /** @var list<array{row: int, col: int, cell: ImageCell}> */
```

src/Writer/Common/Entity/Worksheet.php
```diff
@@ -81,6 +86,19 @@ public function setLastWrittenRowIndex(int $lastWrittenRowIndex): void
         $this->lastWrittenRowIndex = $lastWrittenRowIndex;
     }
 
+    public function addImage(int $row, int $col, ImageCell $cell): void
+    {
+        $this->images[] = ['row' => $row, 'col' => $col, 'cell' => $cell];
+    }
+
+    /**
+     * @return array<array{row: int, col: int, cell: ImageCell}>
```
Slamdunk:
```suggestion
     * @return list<array{row: int, col: int, cell: ImageCell}>
```

tests/Writer/ODS/WriterTest.php
```diff
@@ -487,6 +489,26 @@ public function testAddAutofilterToTwoSheetsShouldWriteCorrectData(): void
...
+            $fileName = 'test_image_cell_throws.ods';
+            $resourcePath = (new TestUsingResource())->getGeneratedResourcePath($fileName);
+            $options = new Options(tempFolder: (new TestUsingResource())->getTempFolderPath());
+            $writer = new Writer($options);
+            $writer->openToFile($resourcePath);
+
+            $this->expectException(UnsupportedTypeException::class);
+            $writer->addRow(new Row([new ImageCell($imagePath, 1, 1)]));
+        } finally {
+            unlink($imagePath);
```
Slamdunk: Not a fun of polluting the indentation just for this: let's get rid of the `unlink` and the `try-catch`

tests/Writer/XLSX/WriterTest.php
```diff
@@ -1448,6 +1449,98 @@ public function testDataValidationsOnSeparateSheets(): void
...
+            self::assertNotFalse($contentTypes);
+            self::assertStringContainsString('drawing+xml', $contentTypes);
+            self::assertStringContainsString('image/png', $contentTypes);
+
+            $sheetRels = file_get_contents('zip://'.$resourcePath.'#xl/worksheets/_rels/sheet1.xml.rels');
+            self::assertNotFalse($sheetRels);
+            self::assertStringContainsString('rIdDrawing1', $sheetRels);
+            self::assertStringContainsString('drawings/drawing1.xml', $sheetRels);
+        } finally {
+            unlink($imagePath);
```
Slamdunk: Same here

tests/Writer/XLSX/WriterTest.php
```diff
@@ -1448,6 +1449,98 @@ public function testDataValidationsOnSeparateSheets(): void
...
+            for ($i = 0; $i < $zip->numFiles; ++$i) {
+                $name = $zip->getNameIndex($i);
+                if (false !== $name && str_starts_with($name, 'xl/media/')) {
+                    ++$mediaCount;
+                }
+            }
+            $zip->close();
+            self::assertSame(1, $mediaCount, 'The same image should be stored only once in the archive');
+        } finally {
+            unlink($imagePath);
```
Slamdunk: Same here

Slamdunk (changes requested):

tests/Writer/ODS/WriterTest.php
thelfensdrfer: > just for this

In this case "this" would be a temporary file not being deleted if the test fails or results in an unexpected error.

I understand why you would find it nicer to look at but why remove the unlink at all? We could at least keep it for the default case, the passing test.

src/Common/Entity/Cell.php
thelfensdrfer: See my example in the description:

```php
$cells = [
    Cell::fromImagePath(
        $path,
        width: $width,
        height: $height,
        fitToCell: true,
    ),
    Cell::fromValue('Carl'),
    Cell::fromValue('is'),
    Cell::fromValue('great!'),
];
```

tests/Writer/ODS/WriterTest.php
Slamdunk: Please use `setUp` and `tearDown` method then

src/Common/Entity/Cell.php
thelfensdrfer: Another way to do it would be to create a new class for images and use it like this for example:

```php
$image = new XlsxImage(
    $path,
    width: $width,
    height: $height,
    fitToCell: true,
);

$cells = [
    Cell::fromValue($image),
    Cell::fromValue('Carl'),
    Cell::fromValue('is'),
    Cell::fromValue('great!'),
];
```

Then we wouldn't need an unused method, but have an unused class.

Slamdunk: `Cell::fromValue` helps the user by guessing the content type.

But with images you already know the content type, so the following is enough:

```php
$cells = [
    new Cell\ImageCell(
        $path,
        width: $width,
        height: $height,
        fitToCell: true,
    ),
    Cell::fromValue('Carl'),
    Cell::fromValue('is'),
    Cell::fromValue('great!'),
];
```

thelfensdrfer: Yeah, that would be another possibility, but isn't it slower to guess the type for every single cell?

Edit: Not just in this instance but in general.

thelfensdrfer: Sry, read your code example wrong. Of course the helper method is slower than using the class directly, but thats what you already did.

Slamdunk: `Cell::fromValue` is slower, but useful.

`Cell::fromImagePath` is slower and useless, compared to directly call the respective class constructor.

thelfensdrfer: Maybe, for me it's easier to autocomplete a method name when I know the class and how the method starts. So `Cell::from*` would autocomplete all methods like `fromImagePath`, `fromDateTime`, etc. But that's not the point here, code is updated.

tests/Writer/CSV/WriterTest.php
```diff
@@ -273,6 +275,25 @@ public function testShouldReturnWrittenRowCount(): void
...
+        try {
+            $fileName = 'test_image_cell_throws.csv';
+            $resourcePath = (new TestUsingResource())->getGeneratedResourcePath($fileName);
+            $writer = new Writer();
+            $writer->openToFile($resourcePath);
+
+            $this->expectException(UnsupportedTypeException::class);
+            $writer->addRow(new Row([new ImageCell($imagePath, 1, 1)]));
+        } finally {
+            unlink($imagePath);
```
Slamdunk: I guess the `ImageHelperTrait` can be used here as well

src/Common/Entity/Cell.php
```diff
@@ -11,6 +11,7 @@
 use OpenSpout\Common\Entity\Cell\DateTimeCell;
 use OpenSpout\Common\Entity\Cell\EmptyCell;
 use OpenSpout\Common\Entity\Cell\FormulaCell;
+use OpenSpout\Common\Entity\Cell\ImageCell;
```
Slamdunk: `use` statement non necessary anymore

Slamdunk (changes requested):

src/Common/Entity/Cell.php
thelfensdrfer: Whoops, normally phpstorm automatically removes unused imports for me, but I disabled everything in this project because of cs fixer. Another unused import is here: `src/Writer/ODS/Helper/BorderHelper.php` (for another pr)

Slamdunk (approved):

thelfensdrfer:
The problem of the failing actions is the following conflict:

  - phpunit/phpunit 13.2.0 needs sebastian/diff ^9.0
  - infection/infection 0.33.2 only supports up to ^8.0

Since infection 0.33 is the most recent package, we would have to pin phpunit to 13.1.*

Slamdunk:
The issue is that @renovatebot self-merged a PR with failing checks: https://github.com/openspout/openspout/pull/387

I'll take care of that, thank you

nikspyratos:
Heyo. Any assistance I could provide for getting this merged in?

Slamdunk:
@thelfensdrfer @nikspyratos I have fixed the base branch, can you rebase the PR please?

nikspyratos:
@Slamdunk https://github.com/openspout/openspout/pull/393

Slamdunk:
Closed by https://github.com/openspout/openspout/pull/393

thelfensdrfer:
> @Slamdunk https://github.com/openspout/openspout/pull/393

Thanks! Didn't have the time in the last few days.

PR #382, opened 2026-05-12
Style deduplication in AbstractStyleRegistry uses object identity instead of object content

michieljmmaas:
### Problem

When generating large XLSX files, Excel warns about exceeding the unique font limit (32,767). When I checked the generated archive I saw thousands of duplicate `<font>` entries in `xl/styles.xml` — in my case ~73,000 entries for only about 10 distinct fonts.

The cause is in `AbstractStyleRegistry`. The deduplication is keyed on `spl_object_hash($style)`, which is based on the object's identity (memory pointer), not its content. So two `Style` objects that look identical but are different instances both get registered, producing duplicate entries in the output.

### Why this is a problem in v5 specifically

This was introduced in [this commit as part of the v5 update](https://github.com/openspout/openspout/pull/338/changes/0f40edd7eba611e04f44dc8f87e6688a9adf966d), probably for performance — `spl_object_hash` is much cheaper than `serialize`. It works fine if you always reuse the same style instance. But v5 also made `Style` immutable, meaning `with*` methods always return a **new instance**. So `getDefaultStyle()->withFontBold(true)` called twice produces two hashes, two registry entries, two `<font>` elements. The two changes conflict with each other.

### Fix

Replace `spl_object_hash($style)` with `serialize($style)` so deduplication is based on the style's content rather than its instance.

`serialize` is significantly slower than `spl_object_hash` — we're talking nanoseconds vs microseconds per call. On a large file that adds up. But without proper deduplication you end up writing thousands of duplicate `<font>` entries to `styles.xml`, which is far more expensive than the overhead of `serialize`. So the performance tradeoff still favors this fix.

I'd still like to hear the original reasoning for the `spl_object_hash` change — maybe there's something I'm missing. But this fix works for my case and seems correct to me.

### DIFF
--- a/tests/Writer/ODS/WriterWithStyleTest.php
+++ b/tests/Writer/ODS/WriterWithStyleTest.php
@@ -300,7 +300,7 @@ final class WriterWithStyleTest extends TestCase
 
         $styleElements = $this->getCellStyleElementsFromContentXmlFile($fileName);
 
-        self::assertCount(5, $styleElements);
+        self::assertCount(4, $styleElements);
 
         // Use reflection for protected members here
         $widthMap = BorderHelper::widthMap;
@@ -328,7 +328,7 @@ final class WriterWithStyleTest extends TestCase
             Color::RED
         );
 
-        $actualThird = $styleElements[3]
+        $actualThird = $styleElements[2]
             ->getElementsByTagName('table-cell-properties')
             ->item(0)
             ->getAttribute('fo:border-top')
--- a/src/Writer/Common/Manager/Style/AbstractStyleRegistry.php
+++ b/src/Writer/Common/Manager/Style/AbstractStyleRegistry.php
@@ -33,7 +33,7 @@ abstract class AbstractStyleRegistry
      */
     final public function registerStyle(Style $style): int
     {
-        $serializedStyle = spl_object_hash($style);
+        $serializedStyle = serialize($style);
         if (\array_key_exists($serializedStyle, $this->serializedStyleToStyleIdMappingTable)) {
             return $this->serializedStyleToStyleIdMappingTable[$serializedStyle];
         }
--- a/tests/Writer/XLSX/Manager/Style/StyleRegistryTest.php
+++ b/tests/Writer/XLSX/Manager/Style/StyleRegistryTest.php
@@ -87,4 +87,18 @@ final class StyleRegistryTest extends TestCase
         self::assertSame(164, $styleRegistry->getFormatIdForStyleId($styleIdUser), 'Second style with user format set should have index 164 (0 is for the default style)');
         self::assertSame(0, $styleRegistry->getFormatIdForStyleId($styleIdNo), 'Style with no format should have index 0');
     }
+
+    public function testDeduplicateEqualStyles(): void
+    {
+        $style_a = new Style(fontBold: true);
+        $styleRegistry = new StyleRegistry($style_a);
+
+        $style_b = new Style(fontBold: true);
+
+        $styleIdA = $styleRegistry->registerStyle($style_a);
+        $styleIdB = $styleRegistry->registerStyle($style_b);
+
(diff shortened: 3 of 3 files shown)


tests/Writer/ODS/WriterWithStyleTest.php
michieljmmaas, line 1: This test had to be fixed because it had the non-deduplication bug that I described. 

```php
        $styles = [
            new Style(border: $borderBottomGreenThickSolid),
            new Style(), // <-- Empty style
            new Style(border: $borderTopRedThinDashed),
            new Style(border: $borderLeft),
        ];
``` 

In the code above here an Style style is registered. That was already provided as the default style to the AbstractStyleRegistry, so it is deduplicated.

Slamdunk:
> It works fine if you always reuse the same style instance. But v5 also made `Style` immutable, meaning `with*` methods always return a **new instance**. So `getDefaultStyle()->withFontBold(true)` called twice produces two hashes, two registry entries, two `<font>` elements. The two changes conflict with each other.

Yeah: you shouldn't call that twice in the first place.
Determinism and reproducibility are key factors in good programming: carefully plan ahead your style instances to be useful and unique, and reuse them accordingly

michieljmmaas:
I see you added the 'invalid' tag. Does that mean this mean you won't include this change? 

I get your point about planning styles ahead of time, but with complex exports and many different styles, it can get unruly fast. I think other people might run into the same unintended behaviour, because the `with` API naturally leads you to create new style instances.

I expect the original change was made to save on performance, as `spl_object_hash` is much faster than `serialize`. But I feel that performance is lost when when writing duplicate data to disk. 

I wanted to test this theory so I created a performance test, using testing both serialization methods and both usages patterns. These were my results. I have added a patch file for the test, and my result.json to this post. 

**Reused instances (your suggested approach):**

| Scenario | `serialize` | `spl_object_hash` |
| :--- | :--- | :--- |
| small, few styles | 0.009s / 8.9KB | 0.010s / 8.9KB |
| small, many styles | 0.008s / 9.3KB | 0.008s / 9.9KB |
| large, few styles | 0.731s / 427.1KB | 0.695s / 427.1KB |
| large, many styles | 0.72s / 455.9KB | 0.719s / 544.3KB |

**New instances (natural `with*` usage):**

| Scenario | `serialize` | `spl_object_hash` |
| :--- | :--- | :--- |
| small, few styles | 0.013s / 9.3KB | 0.018s / 13.4KB |
| small, many styles | 0.013s / 9.3KB | 0.017s / 13.4KB |
| large, few styles | 1.211s / 455.9KB | 1.816s / 860.4KB |
| large, many styles | 1.169s / 455.9KB | 1.787s / 860.5KB |

When styles are reused, both implementations perform virtually identically, the difference is up to variance. But when new instances are created, `spl_object_hash` is significantly slower and produces files nearly twice the file size on large exports.

So `serialize` seems to have hardly any downsides (even in your preferred usage pattern) and saves time and filespaces in many cases.

Interestingly, in the same commit that switched from `serialize` to `spl_object_hash`, the `testBorders` assertion was also changed from `4` to `5` — which suggests even the tests were adjusted to match the new behaviour rather than the intended one.

If you're set on keeping `spl_object_hash` I'll refactor my own code to work around it. But I do think this is worth reconsidering.

[Serialization_Performance_Test.patch](https://github.com/user-attachments/files/27706427/Serialization_Performance_Test.patch)

[openspout_performance_results.json](https://github.com/user-attachments/files/27706422/openspout_performance_results.json)

Slamdunk:
`serialize` has any downsides... for how the library is coded _right now_.
In the past it was the main culprit for segfaults, is very easy to get it wrong again in the future and hard to track its impact.

> I get your point about planning styles ahead of time, but with complex exports and many different styles, it can get unruly fast

It's not up to this library to solve complexity on user's side.
I could tell you to buy a faster computer, with the same reasoning.
My intention is not to be rude, but to underline how important is to fix complexity instead of asking the tools the complexity run on to get faster.

PR #386, opened 2026-06-04
Add internal SheetViewInterface abstraction and extend sheet view support to ODS

WCYa:
### Summary

Introduce a SheetViewInterface to abstract sheet view serialization
for different spreadsheet formats.

### Changes

- Extract SheetViewInterface from existing XLSX SheetView implementation
- Add ODS ViewSettings implementation
- Both formats now follow a unified serialization contract (getXml)

### Tests

- Added  regression tests for ODS view settings serialization
- Verified settings.xml output correctness

### DIFF
--- a/src/Writer/Common/Entity/SheetViewInterface.php
+++ b/src/Writer/Common/Entity/SheetViewInterface.php
@@ -12,7 +12,7 @@ interface SheetViewInterface
      * The file format layer is responsible for persisting this value into
      * the target document structure.
      *
-     * @return string Format-specific representation of sheet view settings
+     * @return non-empty-string Format-specific representation of sheet view settings
      */
     public function getXml(): string;
 }
--- a/src/Writer/Common/Entity/SheetViewInterface.php
+++ b/src/Writer/Common/Entity/SheetViewInterface.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\Common\Entity;
+
+interface SheetViewInterface
+{
+    /**
+     * Returns the format-specific serialization of the sheet view configuration.
+     *
+     * The file format layer is responsible for persisting this value into
+     * the target document structure.
+     *
+     * @return string Format-specific representation of sheet view settings
+     */
+    public function getXml(): string;
+}
--- a/src/Writer/ODS/Entity/ViewSettings.php
+++ b/src/Writer/ODS/Entity/ViewSettings.php
@@ -4,36 +4,21 @@ declare(strict_types=1);
 
 namespace OpenSpout\Writer\ODS\Entity;
 
-use OpenSpout\Common\Exception\InvalidArgumentException;
 use OpenSpout\Writer\Common\Entity\SheetViewInterface;
 
-class ViewSettings implements SheetViewInterface
+final readonly class ViewSettings implements SheetViewInterface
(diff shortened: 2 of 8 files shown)


src/Writer/Common/Entity/SheetViewInterface.php
```diff
@@ -0,0 +1,18 @@
...
+
+interface SheetViewInterface
+{
+    /**
+     * Returns the format-specific serialization of the sheet view configuration.
+     *
+     * The file format layer is responsible for persisting this value into
+     * the target document structure.
+     *
+     * @return string Format-specific representation of sheet view settings
```
Slamdunk:
```suggestion
     * @return non-empty-string Format-specific representation of sheet view settings
```

src/Writer/ODS/Entity/ViewSettings.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenSpout\Writer\ODS\Entity;
+
+use OpenSpout\Common\Exception\InvalidArgumentException;
+use OpenSpout\Writer\Common\Entity\SheetViewInterface;
+
+class ViewSettings implements SheetViewInterface
```
Slamdunk:
```suggestion
final readonly class ViewSettings implements SheetViewInterface
```

src/Writer/ODS/Entity/ViewSettings.php
```diff
@@ -0,0 +1,79 @@
...
+    ) {
+        if ($freezeRow < 0) {
+            throw new InvalidArgumentException('Freeze row must be a non-negative integer');
+        }
+        $this->freezeRow = $freezeRow;
+
+        if ($freezeColumn < 0) {
+            throw new InvalidArgumentException('Freeze column must be a non-negative integer');
+        }
+        $this->freezeColumn = $freezeColumn;
```
Slamdunk:
```suggestion
    /**
     * @param non-empty-string $sheetName
     * @param non-negative-int $freezeRow
     * @param non-negative-int $freezeColumn
     *
     * @throws InvalidArgumentException
     * 
     */
    public function __construct(
        private string $sheetName,
        public int $freezeRow = 0,
        public int $freezeColumn = 0,
        public bool $showGrid = true,
    ) {
        if ($this->freezeRow < 0) {
            throw new InvalidArgumentException('Freeze row must be a non-negative integer');
        }

        if ($this->freezeColumn < 0) {
            throw new InvalidArgumentException('Freeze column must be a non-negative integer');
        }
```

Slamdunk (changes requested):

src/Writer/ODS/Entity/ViewSettings.php
WCYa: Thanks, I've removed the runtime check since the parameter is already documented as `non-negative-int`.

src/Writer/ODS/Entity/ViewSettings.php
WCYa: Updated, thanks!

src/Writer/Common/Entity/SheetViewInterface.php
WCYa: Updated, thanks!

Slamdunk (approved):

PR #395, opened 2026-07-23
File handle leak when adding multiple sheets via `addNewSheetAndMakeItCurrent()`

vermaysha:
**Describe the bug**
Every call to addNewSheetAndMakeItCurrent() opens 3 file handles (sheet XML, comments XML, VML drawing) and never closes them until Writer::close(). When many sheets are added this quickly exhausts the OS file descriptor limit, causing EMFILE ("Too many open files").

**Environment**
OpenSpout version: v5.8
PHP: 8.4
OS: Linux (any OS with per-process FD limits; default 1024 on Linux)

**To reproduce**
```bash
git clone https://github.com/vermaysha/openspout-file-description-overflow-repro
cd openspout-file-description-overflow-repro
composer install
php ./src/Repro.test or composer run test
```

Or you can run this code on your existing project

```php
<?php
require_once __DIR__ . '/vendor/autoload.php';

$countFds = function (): int {
    $fds = scandir('/proc/self/fd');
    \assert(false !== $fds);
    return \count($fds) - 2;
};

$writer = new Writer();
$writer->openToFile('/tmp/fd-leak-test.xlsx');

$initialFds = $countFds();
echo "FD after openToFile: {$initialFds}\n";

for ($i = 0; $i < 20; $i++) {
    $writer->addNewSheetAndMakeItCurrent();
    printf("Sheet %2d → FDs: %d  (Δ: %+d)\n",
        $i + 1, $countFds(), $countFds() - $initialFds);
}

echo "Before close(): {$countFds()}\n";
$writer->close();
echo "After close():  {$countFds()}\n";
```

**Expected behavior (after fix)**
```
FD after openToFile: 6
Sheet  1 → FDs: 6  (Δ: +0)
Sheet  2 → FDs: 6  (Δ: +0)
Sheet  3 → FDs: 6  (Δ: +0)
...
Sheet 20 → FDs: 6  (Δ: +0)
Before close(): 6
After close():  4
```

**Actual behavior (before fix)**
```
FD after openToFile: 6
Sheet  1 → FDs: 9  (Δ: +3)
Sheet  2 → FDs: 9  (Δ: +3)
Sheet  3 → FDs: 12  (Δ: +6)
Sheet  4 → FDs: 15  (Δ: +9)
Sheet  5 → FDs: 18  (Δ: +12)
...
Sheet 20 → FDs: 63  (Δ: +57)
Before close(): 63
After close():  4
```

Each new sheet adds +3 FDs, only reclaimed on Writer::close(). With the default 1024 FD limit, a crash occurs at ~340 sheets (1024 ÷ 3).

**Root cause**
`AbstractWorkbookManager::addNewSheet()` calls `WorksheetManager::startSheet()` which opens 3 file pointers (sheet XML, comments XML, VML drawing). When switching sheets via `setCurrentSheet()`/`addNewSheetAndMakeItCurrent()`, the old sheet's handles remain open. Only `Writer::close()` → `closeAllWorksheets()` eventually closes them.

Fix
Introduce `closeSheet()`/`resumeSheet()` on `WorksheetManagerInterface`:
1. On sheet switch, close previous sheet's handles without writing XML footers
2. Reopen target sheet's handles in append mode
3. 6 files modified (interface, XLSX & ODS worksheet managers, comments manager, workbook manager)

**The test cases have been created, and all of them have passed.**

### DIFF
--- a/src/Writer/Common/Manager/WorksheetManagerInterface.php
+++ b/src/Writer/Common/Manager/WorksheetManagerInterface.php
@@ -40,10 +40,11 @@ interface WorksheetManagerInterface
     public function close(Worksheet $worksheet): void;
 
     /**
-     * Closes the worksheet's file handles without writing footers.
+     * Suspends the worksheet's file handles without writing footers.
      * Used when switching away from a sheet to prevent file handle accumulation.
+     * Counterpart to resumeSheet().
      */
-    public function closeSheet(Worksheet $worksheet): void;
+    public function suspendSheet(Worksheet $worksheet): void;
 
     /**
      * Reopens the worksheet's file handles in append mode.
--- a/src/Writer/Common/Manager/WorksheetManagerInterface.php
+++ b/src/Writer/Common/Manager/WorksheetManagerInterface.php
@@ -38,4 +38,16 @@ interface WorksheetManagerInterface
      * Closes the worksheet.
      */
     public function close(Worksheet $worksheet): void;
+
+    /**
+     * Closes the worksheet's file handles without writing footers.
+     * Used when switching away from a sheet to prevent file handle accumulation.
+     */
+    public function closeSheet(Worksheet $worksheet): void;
+
+    /**
+     * Reopens the worksheet's file handles in append mode.
+     * Used when switching back to a previously closed sheet.
+     */
+    public function resumeSheet(Worksheet $worksheet): void;
 }
--- a/src/Writer/Common/Entity/Worksheet.php
+++ b/src/Writer/Common/Entity/Worksheet.php
@@ -15,7 +15,7 @@ final class Worksheet
     private readonly string $filePath;
 
     /** @var null|resource Pointer to the sheet data file (e.g. xl/worksheets/sheet1.xml) */
-    private $filePointer = null;
+    private $filePointer;
 
     /** @var Sheet The "external" sheet */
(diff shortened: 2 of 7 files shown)


src/Writer/Common/Manager/WorksheetManagerInterface.php
```diff
@@ -38,4 +38,16 @@ public function startSheet(Worksheet $worksheet): void;
      * Closes the worksheet.
      */
     public function close(Worksheet $worksheet): void;
+
+    /**
+     * Closes the worksheet's file handles without writing footers.
+     * Used when switching away from a sheet to prevent file handle accumulation.
+     */
+    public function closeSheet(Worksheet $worksheet): void;
```
Slamdunk: The naming here is confusing, there is already a `close(Worksheet $worksheet): void` API.
Could you be more specific here please?

Slamdunk (changes requested):

src/Writer/Common/Manager/WorksheetManagerInterface.php
vermaysha: Thanks for the review. After thinking about it some more, I realize the name is a bit confusing. I’ll change it to `suspendSheet()` since it’s the opposite of `resumeSheet()`.

Slamdunk (approved):

PR #400, opened 2026-08-03
Fix quadratic row building in the ODS reader

### LINKED ISSUE #399: ODS reader: quadratic row building makes number-columns-repeated files unusably slow
### Summary

The ODS reader spends quadratic time expanding `table:number-columns-repeated`. A 19 KB spreadsheet with 17 rows and 9 real columns – a plain LibreOffice export – took **28.5 s** to read. Hoisting one array out of two loops takes it to **0.05 s** with byte-identical output.

The cause is [#338](https://github.com/openspout/openspout/pull/338), which made `Row` `readonly` and would by itself be a performance and design improvement. The two repeat-expansion loops in `Reader/ODS/RowIterator.php` were translated mechanically from `$row->addCell($cell)` (an O(1) in-place append) into copy-array + `withCells()` (an O(n) copy **plus** an O(n) re-validation in the `Row` constructor), while staying inside the loop. So appending N repeated cells costs O(N²) instead of O(N).

### Environment

- PHP 8.5.8, macOS. Measured on both openspout 5.3.0 (where we hit it) and 5.8.0.
- The offending loops are byte-identical in `v5.8.0` – the patch below applies cleanly to both – and `Row::__construct()` gained a third per-cell check via [#370](https://github.com/openspout/openspout/pull/370). 5.8.0 measures **10.9 % slower** than 5.3.0 on this file, consistent with that extra check.
- Timings are the mean of the middle three of seven runs per configuration, discarding the two fastest and two slowest.

### Reproduction
Self-contained, no fixture file needed. Builds a minimal ODS with 17 rows of 9 real cells plus a trailing cell repeated 16 365 times (16 374 columns total):

```php
<?php
require 'vendor/autoload.php';

$rows = '';
for ($r = 0; $r < 17; ++$r) {
    $rows .= '<table:table-row>';
    for ($c = 0; $c < 9; ++$c) {
        $rows .= '<table:table-cell office:value-type="string"><text:p>v</text:p></table:table-cell>';
    }
    $rows .= '<table:table-cell table:number-columns-repeated="16365"/></table:table-row>';
}
$content = '<?xml version="1.0" encoding="UTF-8"?>'
    .'<office:document-content'
    .' xmlns:office="urn:oasis:names:tc:opendocument:xmlns:office:1.0"'
    .' xmlns:table="urn:oasis:names:tc:opendocument:xmlns:table:1.0"'
    .' xmlns:text="urn:oasis:names:tc:opendocument:xmlns:text:1.0"'
    .' xmlns:style="urn:oasis:names:tc:opendocument:xmlns:style:1.0" office:version="1.2">'
    .'<office:automatic-styles>'
    .'<style:style style:name="ta1" style:family="table">'
    .'<style:table-properties table:display="true"/></style:style>'
    .'</office:automatic-styles>'
    .'<office:body><office:spreadsheet><table:table table:name="Sheet1" table:style-name="ta1">'
    .$rows.'</table:table></office:spreadsheet></office:body></office:document-content>';

$file = sys_get_temp_dir().'/openspout-repro.ods';
@unlink($file);
$zip = new ZipArchive();
$zip->open($file, ZipArchive::CREATE);
$zip->addFromString('mimetype', 'application/vnd.oasis.opendocument.spreadsheet');
$zip->addFromString('META-INF/manifest.xml',
    '<?xml version="1.0" encoding="UTF-8"?><manifest:manifest'
    .' xmlns:manifest="urn:oasis:names:tc:opendocument:xmlns:manifest:1.0" manifest:version="1.2">'
    .'<manifest:file-entry manifest:full-path="/" manifest:media-type="application/vnd.oasis.opendocument.spreadsheet"/>'
    .'<manifest:file-entry manifest:full-path="content.xml" manifest:media-type="text/xml"/>'
    .'</manifest:manifest>');
$zip->addFromString('content.xml', $content);
$zip->close();

$t = microtime(true);
$reader = new OpenSpout\Reader\ODS\Reader();
$reader->open($file);
$rowCount = 0; $cellCount = 0;
    foreach ($sheet->getRowIterator() as $row) { ++$rowCount; $cellCount += \count($row->cells); }
    break;
}
$reader->close();
printf("rows=%d cells=%d time=%.2fs\n", $rowCount, $cellCount, microtime(true) - $t);
```

Output for the **974-byte** generated file:

| openspout | time |
|---|---|
| 5.3.0 | 29.4 s |
| 5.8.0 | 32.6 s |
| 5.8.0 + the fix below | **0.03 s** |

All three return the same 17 rows / 278 358 cells.

### Root cause

`src/Reader/ODS/RowIterator.php`, `processCellStartingNode()` (~line 225) and `processRowEndingNode()` (~line 263):

```php
for ($i = 0; $i < $actualNumColumnsRepeated; ++$i) {
    $cells = $this->currentlyProcessedRow->cells;                                  // O(n) copy
    $cells[] = $this->lastProcessedCell;
    $this->currentlyProcessedRow = $this->currentlyProcessedRow->withCells($cells); // O(n) validation
}
```

Both operations inside the loop are linear in the number of cells accumulated so far, so expanding a repeat of N costs O(N²). With N = 16 365 that is ~134 M array element copies and 16 365 `Row` allocations *per row*.

Note the guard just above the second loop does not save this file:

```php
if (($numCellsInCurrentlyProcessedRow + $actualNumColumnsRepeated) !== self::MAX_COLUMNS_EXCEL) {
```

It only skips the trailing filler when the row totals exactly 16 384 columns. LibreOffice writes 16 374 here (9 + 16 365), so the filler is materialised in full. That behaviour predates v5 and is arguably a second, separate issue – v4 built the same cells, just via O(1) appends – but it is what turns the quadratic loop into a hard failure rather than a mild slowdown.

### Proposed fix

Hoist the array out of both loops so the `Row` is rebuilt once instead of once per repeated cell:

```diff
-            for ($i = 0; $i < $actualNumColumnsRepeated; ++$i) {
-                $cells = $this->currentlyProcessedRow->cells;
-                $cells[] = $this->lastProcessedCell;
-                $this->currentlyProcessedRow = $this->currentlyProcessedRow->withCells($cells);
-            }
+            $cells = $this->currentlyProcessedRow->cells;
+            for ($i = 0; $i < $actualNumColumnsRepeated; ++$i) {
+                $cells[] = $this->lastProcessedCell;
+            }
+            $this->currentlyProcessedRow = $this->currentlyProcessedRow->withCells($cells);
```

Same change in both places – the above is `processRowEndingNode()`; the loop in `processCellStartingNode()` is identical except that it counts on `$this->numColumnsRepeated`. With this applied the repro drops from **32.6 s to 0.03 s** on 5.8.0, with byte-identical output (17 rows, 278 358 cells).

### The same pattern is in the XLSX reader

`src/Reader/XLSX/RowIterator.php:325-327` rebuilds the whole `Row` once per cell:

```php
$cells = $this->currentlyProcessedRow->cells;
$cells[$currentColumnIndex] = $cell;
$this->currentlyProcessedRow = $this->currentlyProcessedRow->withCells($cells);
```

Because line 305 pre-fills the array to the full declared row width, the cost per row is (cells written × declared width), i.e. quadratic in row width. Measured with a constant 800 000 cells per file, comparing 5.3.0 against a variant that accumulates in a plain array and builds the `Row` once in `processRowEndingNode()`:

| row width | 5.3.0 | Row built once | overhead |
|---|---|---|---|
| 20 cols × 40 000 rows | 4.32 s | 4.02 s | +7 % |
| 100 cols × 8 000 rows | 5.11 s | 3.96 s | +29 % |
| 400 cols × 2 000 rows | 8.16 s | 3.97 s | +106 % |
| 1600 cols × 500 rows | 20.29 s | 4.02 s | +404 % |

Narrow sheets are fine, so this is much less urgent than the ODS case. It does become severe when `$numberOfColumnsForRow` is large: a sheet declaring `<dimension ref="A1:XFD500"/>` with no per-row `spans` and 200 real cells per row (356 KB file) reads in **20.8 s** vs **0.64 s** – a 32× improvement.

jhattara:
The two `table:number-columns-repeated` expansion loops in `Reader/ODS/RowIterator.php` copy the cells array and rebuild the `Row` on every appended cell, so expanding a repeat of N costs O(N²) instead of O(N).

This came in with #338, where `$row->addCell($cell)` – an O(1) in-place append – was translated into copy-array + `withCells()` while staying inside the loop. Making `Row` readonly is a clear improvement; this looks like collateral damage in one file rather than anything intended.

A 19 KB LibreOffice export with 17 rows and 9 real columns takes ~29 s to read on 5.3.0 and ~33 s on 5.8.0. Hoisting the array out of both loops takes it to 0.03 s, with byte-identical output.

### Changes

- **fix**: hoist the array out of both repeat loops, so the `Row` is rebuilt once per row instead of once per repeated cell.
- **test**: add `benchReading5RowsODSWithNumberColumnsRepeated` and a 916-byte fixture.

### Benchmark

The fixture is 5 rows of 9 cells, each followed by a trailing empty cell repeated 16 365 times. The 16 374 column total sits just below `MAX_COLUMNS_EXCEL`, so the trailing-filler guard does not skip it – that is the shape a LibreOffice export actually produces.

| reader | time |
|---|---|
| before | ~9.4 s, fails `mode(variant.time.avg) < 1000000` |
| after | ~0.007 s |

It runs under the existing `OdsReaderBench` entry in the benchmark matrix, so no CI change was needed.

### Notes

Peak memory also drops by one array copy per row – 265 920 bytes on this fixture. The old loop's `$cells = $this->currentlyProcessedRow->cells` followed by `$cells[] =` triggers copy-on-write on every iteration, so two full-size arrays were live at once.

`Reader/XLSX/RowIterator.php` has the same pattern in `processCellStartingNode()`, quadratic in row width rather than in repeat count and so far less severe. It is left out of this PR because the impact is much lower and the fix more complicated due to the different method structure: the copy happens once per cell across separate callbacks rather than inside a loop, so it needs the row's cells to accumulate in a plain array and the `Row` to be built once in `processRowEndingNode()`.

Refs #399 – this covers the ODS half of that issue.


Slamdunk (approved):
Awesome, thank you 💪

PR #402, opened 2026-08-05
Fix: truncated comments.xml and vmlDrawing.vml on suspended sheets

### LINKED ISSUE #401: Multi-sheet XLSX files are written with a truncated comments*.xml and vmlDrawing*.vml (v5.10.1)
# Multi-sheet XLSX files are written with a truncated `comments*.xml` and `vmlDrawing*.vml` (v5.10.1)

**Affected:** v5.10.1 · **Works:** v5.10.0 and earlier · **PHP:** 8.4

## Summary

Since v5.10.1, any XLSX with more than one sheet is written with an unclosed `xl/comments{N}.xml`
and `xl/drawings/vmlDrawing{N}.vml` for **every sheet that is not the current one when `close()` is
called**. Both parts are always referenced from the sheet's `.rels`, so the resulting package is
invalid OOXML.

This happens with no comments anywhere in the file, and without ever switching back to a previous
sheet — just adding a second sheet is enough.

Excel refuses the file and offers to repair it. LibreOffice, WPS and Google Drive open it without
complaining, which makes the problem easy to miss.

<details>
  <summary>Images of Microsoft Excel, online version</summary>

  
Image

Image
</details>

## Reproduction

```php

$writer = new Writer();
$writer->openToFile($path);

$writer->getCurrentSheet()->setName('First');
$writer->addRow(Row::fromValues(['hello']));

$writer->addNewSheetAndMakeItCurrent()->setName('Second');
$writer->addRow(Row::fromValues(['world']));

$writer->close();
```

`xl/comments1.xml` in the produced file:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<comments xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
    <authors><author>Unknown</author></authors>
    <commentList>
```

`</commentList>` and `</comments>` are missing. `xl/drawings/vmlDrawing1.vml` is missing its
`</xml>` in the same way. `xl/comments2.xml` and `vmlDrawing2.vml` (the sheet that was current at
`close()`) are complete.

Checking every XML part of the package:

```php
$zip = new ZipArchive();
$zip->open($path);
libxml_use_internal_errors(true);

for ($i = 0; $i < $zip->numFiles; $i++) {
    $name = $zip->getNameIndex($i);
    if (str_ends_with($name, '.xml') && simplexml_load_string($zip->getFromName($name)) === false) {
        echo "MALFORMED: {$name}\n";
    }
}
```

v5.10.1 prints `MALFORMED: xl/comments1.xml`; v5.10.0 prints nothing.

With N sheets, N-1 of them are affected.

[issue-openspout-5.10.0-ok.xlsx](https://github.com/user-attachments/files/30715753/issue-openspout-5.10.0-ok.xlsx)

[issue-openspout-5.10.1-broken.xlsx](https://github.com/user-attachments/files/30715759/issue-openspout-5.10.1-broken.xlsx)

## Cause

Introduced by #395, which added `suspendSheet()`/`resumeSheet()` to limit the number of open file
handles.

`AbstractWorkbookManager::setCurrentWorksheet()` suspends the previous sheet — and
`addNewSheetAndMakeItCurrent()` goes through it, so this is not limited to switching back and forth
between sheets:

```php
private function setCurrentWorksheet(Worksheet $worksheet): void
{
    if (isset($this->currentWorksheet) && $this->currentWorksheet !== $worksheet) {
        $this->worksheetManager->suspendSheet($this->currentWorksheet);
    }
    $this->worksheetManager->resumeSheet($worksheet);
    $this->currentWorksheet = $worksheet;
}
```

`XLSX\Manager\WorksheetManager::suspendSheet()` calls
`CommentsManager::closeTempCommentFiles()`, which closes both files **without writing the footers**
and then unsets the pointers:

```php
public function closeTempCommentFiles(Worksheet $sheet): void
{
    $sheetId = $sheet->getId();

    if (!isset($this->commentsFilePointers[$sheetId])) {
        return;
    }

    fclose($this->commentsFilePointers[$sheetId]);
    fclose($this->drawingFilePointers[$sheetId]);

    unset($this->commentsFilePointers[$sheetId], $this->drawingFilePointers[$sheetId]);
}
```

At the end, `AbstractWorkbookManager::closeAllWorksheets()` calls `WorksheetManager::close()` on
every worksheet, which delegates to `CommentsManager::closeWorksheetCommentFiles()` — the method
that would write the footers. It returns early, precisely because the pointer was unset by the
suspend:

```php
public function closeWorksheetCommentFiles(Worksheet $sheet): void
{
    $sheetId = $sheet->getId();

    if (!isset($this->commentsFilePointers[$sheetId])) {
        return;                                    // <- suspended sheets exit here
    }

    $commentFp = $this->commentsFilePointers[$sheetId];
    $drawingFp = $this->drawingFilePointers[$sheetId];

    fwrite($commentFp, self::COMMENTS_XML_FILE_FOOTER);
    fwrite($drawingFp, self::DRAWINGS_VML_FILE_FOOTER);
    // ...
}
```

Suspended sheets are never resumed before the workbook closes, so the footers are never written.

What makes it reach every user, and not only the ones using comments, is that
`XLSX\Helper\FileSystemHelper` emits the relationship unconditionally, for every sheet:

```php
// Add the legacy drawing for comments
fwrite($worksheetFilePointer, '<legacyDrawing r:id="rId_comments_vml1"/>');
```

so the malformed parts are always referenced from `xl/worksheets/_rels/sheet{N}.xml.rels`.

---

PS: I was implementing a new feature using this lib with Claude Code's help and claude found this problem. We then validated it and claude wrote this issue. I dont know this lib enough to create a PR, but i hope this helps 👍

vermaysha:
## Multi-sheet XLSX files are written with a truncated `comments*.xml` and `vmlDrawing*.vml`

**Affects:** 5.x (since v5.10.1) · **Closes:** #401

### Problem
Introduced by #395, `suspendSheet()` calls `closeTempCommentFiles()` which closes and unsets file pointers **without writing XML footers** (`</commentList></comments>` and `</xml>`). At close time, `closeWorksheetCommentFiles()` sees the pointers unset and returns early-leaving `commentsN.xml` and `vmlDrawingN.vml` truncated for every sheet that is not the current one when `close()` is called.

Any XLSX with 2+ sheets is affected, even without comments.

### Fix
**`CommentsManager::closeWorksheetCommentFiles`** instead of returning early when pointers are unset, reopen the temp files in append mode via `reopenTempCommentFiles()`, then write footers and close normally.

**`CommentsManager::reopenTempCommentFiles`** add `isset` guard to prevent FD leak when `resumeSheet()` is called on a freshly created sheet (pointers already open from `createWorksheetCommentFiles`).

### Tests added
- `testMultiSheetXmlFilesAreWellFormed` 3 sheets, close, validate all XML/VML parts via `simplexml_load_string`
- `testSwitchBetweenSheetsKeepsXmlWellFormed` Sheet1 -> Sheet2 -> Sheet1 -> Sheet2, close, validate all parts well-formed
- All 193 existing XLSX writer tests continue to pass

### DIFF
--- a/tests/Writer/XLSX/WriterTest.php
+++ b/tests/Writer/XLSX/WriterTest.php
@@ -1746,4 +1746,70 @@ final class WriterTest extends TestCase
             'Expected all XML/VML parts well-formed, malformed: '.implode(', ', $malformed)
         );
     }
+
+    /**
+     * Regression: switching between sheets back and forth must keep all XML parts well-formed.
+     *
+     * @see https://github.com/openspout/openspout/issues/401
+     */
+    public function testSwitchBetweenSheetsKeepsXmlWellFormed(): void
+    {
+        $fileName = 'test_switch_sheets_xml_wellformed.xlsx';
+        $resourcePath = (new TestUsingResource())->getGeneratedResourcePath($fileName);
+
+        $options = new Options(tempFolder: (new TestUsingResource())->getTempFolderPath());
+        $writer = new Writer($options);
+        $writer->openToFile($resourcePath);
+
+        $sheet1 = $writer->getCurrentSheet();
+        $sheet1->setName('First');
+        $writer->addRow(Row::fromValues(['sheet1-a']));
+
+        $writer->addNewSheetAndMakeItCurrent()->setName('Second');
+        $writer->addRow(Row::fromValues(['sheet2-a']));
+
+        // switch back to sheet 1, resume then suspend again
+        $writer->setCurrentSheet($sheet1);
+        $writer->addRow(Row::fromValues(['sheet1-b']));
+
+        // switch to sheet 2, resume then suspend again
+        $sheet2 = $writer->getSheets()[1];
+        $writer->setCurrentSheet($sheet2);
+        $writer->addRow(Row::fromValues(['sheet2-b']));
+
+        $writer->close();
+
+        // sheet2 was current at close -> sheet1 suspended, sheet2 normal
+        // sheet1 was suspended twice (once at first switch, once again at second switch)
+        // sheet2 was suspended once (when switching back to sheet1)
+
+        $zip = new ZipArchive();
+        $openResult = $zip->open($resourcePath);
+        self::assertTrue($openResult);
+
+        libxml_use_internal_errors(true);
+        $malformed = [];
+
+        for ($i = 0; $i < $zip->numFiles; ++$i) {
+            $name = $zip->getNameIndex($i);
+            if (!\is_string($name) || !str_ends_with($name, '.xml') && !str_ends_with($name, '.vml')) {
+                continue;
+            }
+
+            $content = $zip->getFromName($name);
+            self::assertNotFalse($content);
+
+            if (false === simplexml_load_string($content)) {
+                $malformed[] = $name;
+            }
+        }
+
+        $zip->close();
+        libxml_use_internal_errors(false);
+
+        self::assertEmpty(
+            $malformed,
+            'Expected all XML/VML parts well-formed after switching sheets back and forth, malformed: '.implode(', ', $malformed)
+        );
+    }
 }
--- a/tests/Writer/XLSX/WriterTest.php
+++ b/tests/Writer/XLSX/WriterTest.php
@@ -1690,4 +1690,60 @@ final class WriterTest extends TestCase
 
         return $xmlReader;
     }
+
+    /**
+     * Regression: multi-sheet XLSX must not produce truncated XML in comments or vmlDrawing files.
+     *
+     * @see https://github.com/openspout/openspout/issues/401
+     */
+    public function testMultiSheetXmlFilesAreWellFormed(): void
+    {
+        $fileName = 'test_multisheet_xml_wellformed.xlsx';
+        $resourcePath = (new TestUsingResource())->getGeneratedResourcePath($fileName);
+
+        $options = new Options(tempFolder: (new TestUsingResource())->getTempFolderPath());
+        $writer = new Writer($options);
+        $writer->openToFile($resourcePath);
+
+        // 3 sheets, only last one active at close — forces suspend on sheets 1 & 2
+        $writer->getCurrentSheet()->setName('First');
+        $writer->addRow(Row::fromValues(['hello']));
+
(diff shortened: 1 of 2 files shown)


vermaysha:
I've tested it in the internal app, and there were no issues when exporting over 600 sheets; they opened safely without any warnings in Excel Online and Excel 2021

tests/Writer/XLSX/WriterTest.php
```diff
@@ -1581,6 +1581,128 @@ public function testAddNewSheetAndMakeItCurrentLeaksFileHandles(): void
         );
     }
 
+    /**
+     * Regression: multi-sheet XLSX must not produce truncated XML in comments or vmlDrawing files.
+     *
+     * @see https://github.com/openspout/openspout/issues/401
+     */
+    public function testMultiSheetXmlFilesAreWellFormed(): void
```
Slamdunk, line 1589: Can we test the actual presence of the closing tags?

Slamdunk (changes requested):

tests/Writer/XLSX/WriterTest.php
vermaysha: It seems possible to check whether closing tags really exist or not,

vermaysha: @Slamdunk 
Based on the official PHP documentation, `simplexml_load_string` validates whether the XML is truly well-formed. [simplexml_load_string](https://www.php.net/manual/en/function.simplexml-load-string.php)

This is an example of when an error occurs.

```
1) OpenSpout\Writer\XLSX\WriterTest::testMultiSheetXmlFilesAreWellFormed
Expected all XML/VML parts well-formed, malformed: xl/drawings/vmlDrawing1.vml, xl/drawings/vmlDrawing2.vml, xl/comments1.xml, xl/comments2.xml
Failed asserting that an array is empty.
```

Do you think we need to run another test to verify that the closing tags are actually present? Perhaps using the `str_contains()` function for the `</comments>`, `</commentList>`, and `</xml>` tags?

Slamdunk (approved):
I run the test against upstream and they indeed fail, so we are ok here 👍

PR #405, opened 2026-08-18
feat: implement OptionsInterface in Options classes

finagin:


Slamdunk (changes requested):
`Options` is a value objects, it just stores values, what's the point of the interface?

finagin:
@Slamdunk 

I'm making a wrapper for this library and I'd like to have strong typing but be able to accept any of the three formats

finagin:
If we add an interface for Writer, it would be logical to add it for Reader for consistency.

Slamdunk:
This library already has strong typing.

Value objects don't have an API, there's no logic in adding one.

PR #406, opened 2026-08-18
Feature/options interface

finagin:
❌ Now
```php
use OpenSpout\Writer\CSV;
use OpenSpout\Writer\ODS;
use OpenSpout\Writer\XLSX;

// ...

public function someMethod(CSV\Options|ODS\Options|XLSX\Options $options)
```

✅ PR
```php
use OpenSpout\Writer\OptionsInterface;

// ...

public function someMethod(OptionsInterface $options)
```


Slamdunk:
Maybe you are missing what a value objects is: no API means no behaviour, which means no contract to abstract between provider and consumer.

PR #409, opened 2026-09-02
Fix XLSX freeze pane for row/column-only freezes

### LINKED ISSUE #375: `SheetView::withFreezeColumn()` generates invalid XLSX pane metadata (`ySplit="-1"`,   `topLeftCell="B0"`) when no freeze row is set
## Summary

Using `OpenSpout\Writer\XLSX\Entity\SheetView::withFreezeColumn('B')` generates invalid
worksheet pane XML in XLSX output.

The resulting pane is:

```xml
<pane xSplit="1" ySplit="-1" topLeftCell="B0" activePane="bottomRight" state="frozen"/>
```
This is invalid because:

- ySplit is negative
- topLeftCell="B0" references row 0, which is not a valid Excel row

Excel opens the file in repair mode and reports repaired "View" records in xl/worksheets/
sheet1.xml.

## Environment

- openspout/openspout: v5.6.0
- PHP: 8.4

## Minimal Reproduction

```php
require 'vendor/autoload.php';

use OpenSpout\Writer\XLSX\Entity\SheetView;

$view = (new SheetView())->withFreezeColumn('B');

echo $view->getXml(), PHP_EOL;
```

Actual output:

```xml
<sheetView ...><pane xSplit="1" ySplit="-1" topLeftCell="B0" activePane="bottomRight"
state="frozen"/></sheetView>
```

## Expected Behavior

Column-only freezing should generate valid pane metadata, for example:

```xml
<pane xSplit="1" topLeftCell="B1" activePane="bottomRight" state="frozen"/>
```

or another valid equivalent.

## Likely Cause

In SheetView::getFreezeCellPaneXml() the pane XML is built from:

- ySplit => $this->freezeRow - 1
- topLeftCell => $this->freezeColumn . $this->freezeRow

When only withFreezeColumn('B') is used, freezeRow remains 0, which produces ySplit = -1 and
topLeftCell = B0.

## Suggested Fix

Handle column-only freeze separately so row coordinates never go below 1, and never emit
negative ySplit values.

ziegenberg:
### Summary
`SheetView` generates invalid `<pane>` metadata when freezing on a single axis. This was originally reported in #375: using `withFreezeColumn('B')` emitted `ySplit="-1"` and `topLeftCell="B0"`, which Excel repairs on open.

The same class of bug also affected row-only and multi-axis freezes, along with an incorrect `activePane`.

### Problem
`SheetView::getFreezeCellPaneXml()` unconditionally emitted both splits:

```php
'ySplit' => $this->freezeRow - 1,
'topLeftCell' => $this->freezeColumn.$this->freezeRow,
'activePane' => 'bottomRight',
```

This produced:
- `ySplit="-1"` / `topLeftCell="B0"` for column-only freezes (`freezeRow` 0 or 1)
- a redundant `xSplit="0"` for row-only freezes
- `activePane="bottomRight"` even for single-axis freezes (should be `topRight`/`bottomLeft`)

### Fix
Only emit a split attribute when that axis is actually frozen, clamp the visible cell to a valid row, and select `activePane` from the frozen axes:

| Scenario | Pane |
|---|---|
| Column-only | `<pane xSplit="1" topLeftCell="B1" activePane="topRight" state="frozen"/>` |
| Row-only (first row) | `<pane ySplit="1" topLeftCell="A2" activePane="bottomLeft" state="frozen"/>` |
| Both | `<pane xSplit="1" ySplit="1" topLeftCell="B2" activePane="bottomRight" state="frozen"/>` |

### Tests
Added regression tests for column-only and row-only freezing (valid splits, `topLeftCell`, and `activePane`), and extended the existing pane test. The row-only test freezes the first row.

### DIFF
--- a/src/Writer/XLSX/Entity/SheetView.php
+++ b/src/Writer/XLSX/Entity/SheetView.php
@@ -225,14 +225,26 @@ final readonly class SheetView implements SheetViewInterface
         }
 
         $columnIndex = CellHelper::getColumnIndexFromCellIndex($this->freezeColumn.'1');
+        // A freeze row of 0 or 1 means no rows are frozen, so the visible cell must be in row 1.
+        $freezeRow = max(1, $this->freezeRow);
 
-        return '<pane'.$this->generateAttributes([
-            'xSplit' => $columnIndex,
-            'ySplit' => $this->freezeRow - 1,
-            'topLeftCell' => $this->freezeColumn.$this->freezeRow,
-            'activePane' => 'bottomRight',
-            'state' => 'frozen',
-        ]).'/>';
+        $attributes = [];
+        if ($columnIndex > 0) {
+            $attributes['xSplit'] = $columnIndex;
+        }
+        if ($freezeRow > 1) {
+            $attributes['ySplit'] = $freezeRow - 1;
+        }
+
+        $attributes['topLeftCell'] = $this->freezeColumn.$freezeRow;
+        $attributes['activePane'] = match (true) {
+            $columnIndex > 0 && $freezeRow > 1 => 'bottomRight',
+            $columnIndex > 0 => 'topRight',
+            default => 'bottomLeft',
+        };
+        $attributes['state'] = 'frozen';
+
+        return '<pane'.$this->generateAttributes($attributes).'/>';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Slamdunk:
Thank you

ziegenberg:
That was quick. I'll add a follow-up PR for the positive `xSplit` assertion in `testFreezingOnlyAColumnShouldGenerateValidPaneMetadata`, 'cause I missed that. And while I'm at it, I will also add the suggestion from Copilot, since it makes sense.

PR #411, opened 2026-09-02
fix: write apostrophes in XLSX sheet names unescaped (and enforce reserved "History" name)

### LINKED ISSUE #299: Error when opening a XSLX file with a sheet name containing a `'`
👋 

# Issue
I encountered a strange behavior when setting a sheet name to a name with a `'` (single quote). Excel shows a window saying that the file must be recovered.

I investigated the issue in the workbook.xml file generated by Openspout

```
<sheet name="test with&#039;quote" sheetId="14" r:id="rIdSheet14" state="visible" />
```

If I replace the `'` by a space, Excel is able to open the file just fine.

```
<sheet name="test with no quote" sheetId="2" r:id="rIdSheet2" state="visible"/>
```
# Investigation

It seems that the encoding of `'` to `&#039;` is done by `openspout/src/Common/Helper/Escaper/XLSX.php`

```
public function escape(string $string): string
    {
        $this->initIfNeeded();

        $escapedString = $this->escapeControlCharacters($string);

        // @NOTE: Using ENT_QUOTES as XML entities ('<', '>', '&') as well as
        //        single/double quotes (for XML attributes) need to be encoded.
        return htmlspecialchars($escapedString, ENT_QUOTES, 'UTF-8');
    }
```

I tried to replace the flag `ENT_QUOTES` by `ENT_XML1 | ENT_SUBSTITUTE` or `ENT_COMPAT` but it doesn't fix the issue 🤔 

I was wondering if you ever encountered this issue ? or if maybe point me a function I forgot to change 🤔

ziegenberg:
Fixes #299: XLSX sheet names containing an apostrophe (`'`) are entity-encoded by the XLSX escaper as `&#039;` in `xl/workbook.xml`. Microsoft Excel writes such names with a literal apostrophe and must repair files where it is encoded, prompting "We found a problem with some content...". This change switches the escaper from `ENT_QUOTES` to `ENT_COMPAT` so the apostrophe is no longer encoded, matching Excel's own output. `<`, `>`, `&` and `"` are still escaped, so the XML remains well-formed.

Sheet-name handling now tracks the restrictions documented by Microsoft in ["Rename a worksheet"](https://support.microsoft.com/en-us/excel/rename-a-worksheet):

- the apostrophe is legal in the middle of a name, but names cannot **begin or end** with one,
- names cannot be blank,
- names cannot exceed 31 characters,
- names cannot contain `\ / ? * : [ ]`,
- names cannot use the reserved word "History".

The reserved "History" name is now rejected (case-insensitively), and the start/end-with-apostrophe restriction is now documented on `Sheet::setName()`, bringing the implementation in line with the docs.

### Tests

- Add integration reproduction test: an apostrophe sheet name is written unaltered (and never as `&#039;`).
- Update the XLSX escaper and writer text-content expectations accordingly.
- Add "History" invalid-name data-provider cases (`History`, `HISTORY`, `history`).

### DIFF
--- a/src/Common/Helper/Escaper/XLSX.php
+++ b/src/Common/Helper/Escaper/XLSX.php
@@ -101,9 +101,15 @@ final readonly class XLSX implements EscaperInterface
     {
         $escapedString = $this->escapeControlCharacters($string);
 
-        // @NOTE: Using ENT_QUOTES as XML entities ('<', '>', '&') as well as
-        //        single/double quotes (for XML attributes) need to be encoded.
-        return htmlspecialchars($escapedString, ENT_QUOTES, 'UTF-8');
+        // @NOTE: Using ENT_COMPAT to encode '<', '>', '&' as well as double
+        //        quotes (needed because this writer emits XML attribute values
+        //        that are double-quoted). An apostrophe does not need escaping
+        //        inside a double-quoted attribute (or in text content), and
+        //        Microsoft Excel writes it literally in sheet names while it
+        //        must repair files where the apostrophe is entity-encoded
+        //        as `&#039;` (see OpenSpout issue #299). Other writers that emit
+        //        single-quoted attribute values would need to escape it.
+        return htmlspecialchars($escapedString, ENT_COMPAT, 'UTF-8');
     }
 
     /**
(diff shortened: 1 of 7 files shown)


ziegenberg:
Looking into these comments now.

ziegenberg:
Added both suggestions. Ready for review.

Slamdunk (approved):

PR #412, opened 2026-09-02
fix: keep auto-height rows auto-fitted in XLSX (customHeight absent)

### LINKED ISSUE #367: Excel Online doesn't respect auto-height rows generated by OpenSpout (works in Desktop Excel)
XLSX files generated with OpenSpout using Row::DEFAULT_HEIGHT work correctly in Excel Desktop, where rows automatically adjust their height. However, Excel Online (including viewers in MS Teams, SharePoint) does not respect this auto-height behavior. As a result, rows may appear hidden (in older OpenSpout versions when setHeight(15) was not explicitly used or setHeight(-1), setHeight(0) used) or display with incorrect heights.

Maybe OpenSpout writes different XLSX XML metadata for auto-height rows. Excel Online appears to be more strict about what XML attributes/elements it requires for auto-height to work.

Expected Behavior:
Files generated with Row::DEFAULT_HEIGHT should display correctly with auto-height in both Excel Desktop and Excel Online without requiring post-processing.

ziegenberg:
**Issue:** [#367](https://github.com/openspout/openspout/issues/367) — XLSX files generated with OpenSpout using `Row::DEFAULT_HEIGHT` don't respect auto-height in Excel Online (and MS Teams / SharePoint viewers), where rows can appear hidden or incorrectly sized, while they work fine in Excel Desktop.

**Root cause:** Contrary to the spec, strict viewers infer auto-height from the *absence* of the `customHeight` attribute on a `<row>` element, not from its value. OpenSpout wrote an explicit `customHeight="0"` on default-height rows (and `customHeight="1"` whenever the `DEFAULT_ROW_HEIGHT` option was set, even without an `ht` value). Excel Online treats any present `customHeight` attribute as a manually defined, fixed height, so auto-fit is disabled.

Per ECMA-376 Part 1 ([§18.3.1.73 `row`](https://www.ecma-international.org/publications-and-standards/standards/ecma-376/)), `customHeight` means *"1 if the row height has been manually set"* (default `false`); an auto-fit row is one with `customHeight ≠ true`, and Excel itself encodes it by omitting the attribute entirely. This was verified empirically in Excel Online: an auto row is emitted as `<row r="1" spans="1:1" ht="30.75">` — no `customHeight` — while a fixed row is `<row r="2" spans="1:1" ht="15" customHeight="1">`.

**Not doing:** OpenSpout does not (and should not) compute fitted row heights — that would require reimplementing Excel's text-layout engine. Viewers calculate the fitted `ht` themselves.

### DIFF
--- a/src/Writer/XLSX/Manager/WorksheetManager.php
+++ b/src/Writer/XLSX/Manager/WorksheetManager.php
@@ -112,8 +112,19 @@ final readonly class WorksheetManager implements WorksheetManagerInterface
         $numCells = $row->getNumCells();
 
         $rowHeight = $row->height;
-        $hasCustomHeight = ($this->options->DEFAULT_ROW_HEIGHT > 0 || $rowHeight > 0) ? '1' : '0';
-        $rowXML = "<row r=\"{$rowIndexOneBased}\" spans=\"1:{$numCells}\" ".($rowHeight > 0 ? "ht=\"{$rowHeight}\" " : '')."customHeight=\"{$hasCustomHeight}\">";
+
+        // A row only receives an explicit height (plus the customHeight flag) when one is
+        // actually set. Auto-height rows (Row::DEFAULT_HEIGHT) must carry neither `ht`
+        // nor `customHeight`, otherwise strict viewers (e.g. Excel Online, Teams/SharePoint)
+        // treat them as fixed-height instead of auto-fitting their content.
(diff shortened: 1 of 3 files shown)


ziegenberg:
How often do you want to release today? 😁

Slamdunk:
Releases are free, and more releases allow end user to be more precise when reporting bugs.

PR #413, opened 2026-09-02
docs: fix outdated "edit existing spreadsheet" guide (#408)

### LINKED ISSUE #408: Doc "Edit an existing spreadsheet" is outdated
The provided code samples for editing an existing spreadsheet are not working as methods `getCellAtIndex` and `setCellAtIndex` are no longer available in version 5. What's the optimal way to edit a spreadsheet? Thanks

ziegenberg:
## What & Why

Resolves #408. The code samples in the *"Edit an existing spreadsheet"* guide used the v4 APIs `Row::getCellAtIndex()` / `Row::setCellAtIndex()`, which no longer exist in v5 and made the example impossible to run.

While reproducing the failure, I found the guide had **two other latent bugs** that prevented it from running at all:

1. `Reader\ODS\Options` (and `Reader\XLSX\Options`) are now `final readonly` with constructor-promoted properties, so the old `$readerOptions->SHOULD_FORMAT_DATES = true;` pattern throws `Cannot modify readonly property`.
2. The guide never called `$reader->open($existingFilePath)` before iterating, so it would throw `ReaderNotOpenedException`.

The same stale `Options` pattern existed in the *"Add data to an existing spreadsheet"* guide, so it is fixed there too for consistency.

### DIFF
--- a/docs/guides/1-add-data-existing-spreadsheet.md
+++ b/docs/guides/1-add-data-existing-spreadsheet.md
@@ -18,9 +18,8 @@ $existingFilePath = 'path/to/orders.xlsx';
 $newFilePath = 'path/to/new-orders.xlsx';
 
 // we need a reader to read the existing file...
-$readerOptions = new Options();
-$readerOptions->SHOULD_FORMAT_DATES = true; // this is to be able to copy dates
-$reader = new Reader($readerOptions);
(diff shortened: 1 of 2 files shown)


ziegenberg:
ready to review.

